PR #2, opened 2014-05-29
Cross build

mfleming:
This branch contains some patches that were necessary to cross build chipsec, i.e. where the machine you want to run the tests on isn't the same as the one you compile the source on. 

Please consider pulling.

If you'd prefer the patch series sent to a mailing list or an email address, just let me know.

We're planning on using chipsec for the luvOS to expand our test coverage to include security related test cases. If you're interested, you can look at how we're using chipsec here,

https://github.com/01org/luv-yocto/tree/mfleming/chipsec

### DIFF
--- a/source/tool/chipsec/helper/linux/Makefile
+++ b/source/tool/chipsec/helper/linux/Makefile
@@ -1,8 +1,8 @@
 INC=-I/usr/include/python2.7
 
 cores: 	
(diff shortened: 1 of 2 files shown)


chipsecintel:
I believe these changes were already included in a prior version. Thanks, Matt, for your help getting this working! :-)

PR #5, opened 2015-07-30
Fixing a small ctypes bug that would not give the correct return value.

grrrrrrrrr:


chipsecintel:
This looks right to me.

PR #6, opened 2015-10-26
Fix TABLE NULL constructor name

tweksteen:


chipsecintel:
Yeah... :-( 

Actually, it might be better to do the whole "NULL table" differently, using a base class. We'll try to get this in with the 1.2.2 release.

PR #7, opened 2015-10-28
Clean up setup.py

tweksteen:


chipsecintel:
Seems like this would break on windows, since the .sys files are not included. I think we have another idea that we can merge in with the rest of 1.2.2.

tweksteen:
I thought the Windows build was done using build/build_exe*.py. At least, it is what's mentioned in the documentation. Since any type of build will require some platform specific driver, would it be worth trying to have one setup.py per supported platform? Or a single setup.py that builds according to the current platform? Either way, build/ and the main setup.py should be merged.

chipsecintel:
I think you're referring to the documentation that describes building an EXE. If that's what you want to do, then that build_exe*.py file would be the way to go. However, you can build the windows driver and boot without signature checking to use without an exe as well. 

In any case, take a look at what we did in the latest version. We tried to include conditional behavior depending on windows or linux environment. Please let me know if this works or if you'd suggest some different changes.

PR #8, opened 2015-10-28
Remove global variables from Linux helper

tweksteen:


chipsecintel:
I think this is ok... just want to do some testing.

PR #9, opened 2015-10-28
License cleaning

tweksteen:
There are two license files at the root of the project: COPYING and LICENSE. It seems that LICENSE is an outdated version of COPYING.


chipsecintel:
When we first released chipsec, we intentionally duplicated the license in these files. The diffs seem to be just formatting/whitespace. I don't think there's any problem with it.

PR #16, opened 2015-12-09
Check that hardware sequencing is enabled before interacting with the flash

tweksteen:
Currently, no check is performed to detect that hardware sequencing is enabled (HSFS.FDV). When such configuration is encountered, although the base and limit values are meaningless, Chipsec still interpret them and try to read/write to the flash.

This patch add a method to the SPI HAL to detect such configuration. A call to this method is added to the beginning of each SPI operation.

By doing so, it is still possible to run 'chipsec_utils.py spi info' to get some information on the flash.

### DIFF
--- a/source/tool/chipsec/hal/spi.py
+++ b/source/tool/chipsec/hal/spi.py
@@ -525,6 +525,13 @@ class SPI:
 
         return cycle_done
 
+    def check_hardware_sequencing(self):
+        # Test if the flash decriptor is valid (and hardware sequencing enabled)
+        hsfsts = self.spi_reg_read( self.hsfs_off, 2 )
+        if not (hsfsts & Cfg.PCH_RCBA_SPI_HSFSTS_FDV):
+            logger().error("HSFS.FDV is 0, hardware sequencing is disabled")
+            raise SpiRuntimeError("Chipset does not support hardware sequencing")
+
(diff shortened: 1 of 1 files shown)


tweksteen:
Does this pull request need to be rewritten?

source/tool/chipsec/hal/spi.py
```diff
@@ -525,6 +525,13 @@ def _send_spi_cycle(self, hsfctl_spi_cycle_cmd, dbc, spi_fla ):
 
         return cycle_done
 
+    def check_hardware_sequencing(self):
+        # Test if the flash decriptor is valid (and hardware sequencing enabled)
+        hsfsts = self.spi_reg_read( self.hsfs_off, 2 )
+        if not (hsfsts & Cfg.PCH_RCBA_SPI_HSFSTS_FDV):
+            logger().error("HSFS.FDV is 0, hardware sequencing is disabled")
```
c7zero: Please change to use read_register[_field] instead to make the code platform independent:

fdv = chipsec.chipset.read_register_field( self.cs, 'HSFS', 'FDV' )
if fdv == 0:
    ... raise exception

tweksteen:
The branch was too outdated. See #86 for updated version.

PR #26, opened 2016-04-13
Add test for CMOS dump

tweksteen:


c7zero:
RTC IO ports 70 - 73 are aliased to ports 74 - 77. The test will be ok now as hal/cmos.py only uses ports 70 - 73 but I'd change condition
    if io_port < 0x70 or io_port > 0x73:
to
    if io_port < 0x70 or io_port > 0x77:
going forward.

PR #30, opened 2016-05-27
Move the util functions to classes

tweksteen:
This pull request moves the util commands to specific classes, inheriting from BaseCommand. It includes:

- Define a new class BaseCommand
- For each utils:
  - Convert the function to a class, using the run() method for its main content
  - Use the attributes instead of global variables: `s/logger()/self.logger`, `s/argv/self.argv` and `s/chipsec_utils._cs/self.cs`
  - Add a requires_driver() method that determines if the OS driver needs to be loaded depending on the arguments and potentially other factors (platform).
  - Clean up some of the import if relevant
- Modify chipsec_utils.py to use the new classes

The diff for each command is rather verbose because of the different indentation level. Using `git diff -w` to ignore the space difference may help to review this request.

The purpose of this request is to clean up how the utils are implemented in order to ease the transition when using Chipsec without drivers.

### DIFF
--- a/source/tool/chipsec/command.py
+++ b/source/tool/chipsec/command.py
@@ -1,0 +1,33 @@
+#!/usr/local/bin/python
+#CHIPSEC: Platform Security Assessment Framework
+#Copyright (c) 2016, Google
+#
+#This program is free software; you can redistribute it and/or
+#modify it under the terms of the GNU General Public License
+#as published by the Free Software Foundation; Version 2.
+#
+#This program is distributed in the hope that it will be useful,
+#but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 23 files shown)


c7zero:
Nice change!
One issue to think about regarding requires_driver function. How do we use it in cases like "chipsec_util uefi" where, for example, "chipsec_util uefi s3bootscript" command requires a driver and "chipsec_util uefi decode" command does not. Basically, we'd need per sub-command checking of requires_driver or we'd have to split commands into multiple..

tweksteen:
This case should be trivial to handle. The classes have a `self.argv` attribute where all the command line arguments are stored. You could use a `if self.argv[2] == 'decode': return False` or similar.

PR #35, opened 2016-06-10
chipsec_km: utilize inode_lock/unlock wrappers for new kernels

ricardon:
As of v4.5, wrappers to lock and unlock the inode's mutex were
introduced. This change is useful to leverage future changes in the
implementation of inode locking. For instance, a new change coming in
Linux v4.7-rc1 changed the i_node locking mechanism from using a mutex
to a rw sempahore.

Signed-off-by: Ricardo Neri <ricardo.neri-calderon@linux.intel.com>


c7zero:
New PR #65 was filed on this. Closing to track in #65

PR #40, opened 2016-06-30
Forward the driver loading request to the helpers

tweksteen:
This pull request forward the request to load the driver from the util command to the helper classes. The goal is to still initialise the helper classes even if the driver not be required. Ultimately, it will give the option to use a non-driver method to perform some of the commands.

The two helper methods `start` and `create` are modified to take an extra argument (start_driver). The use and implementation of these is different from each platform. (These should probably be merged in a future commit, since the only calling place uses both one after the other).

The only low-level call required to initialise the driver is reading the platform VID/DID via PCI. This should be possible via userland-only methods on both Linux and Windows. For now, a dummy value is returned '0xFFFF' which will trigger an unknown_platform resolution. The command execution will then continue normally.

### DIFF
--- a/source/tool/chipsec/chipset.py
+++ b/source/tool/chipsec/chipset.py
@@ -248,10 +248,10 @@ class Chipset:
         #   spi = SPI( chipsec_util._cs )
         #
 
-    def init( self, platform_code, start_svc ):
+    def init( self, platform_code, start_driver ):
 
         _unknown_platform = False
-        if start_svc: self.helper.start()
+        self.helper.start(start_driver)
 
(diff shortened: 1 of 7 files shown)


c7zero:
I think, long term, it would be nice to add higher level function identify_platform() which would check DID/VID (among other things like CPUID) and would check if self.driver_loaded before reading DID/VID. Not ideal to return 0xFFFF (which is a valid PCI response) but not expecting any module to call PCI functions if driver is not loaded so should be OK for now.

tweksteen:
Agreed.

PR #51, opened 2016-07-25
Add BOOT_MODE to hardware tests

tweksteen:
Add the BOOT_MODE attribute for hardware tests to probe and filter the platform depending on the boot mode: UEFI or Legacy.


c7zero:
Suggest renaming "bios_version" to "firmware_version"

PR #52, opened 2016-07-25
Add support for /dev/mem to the Linux driver

tweksteen:
This request add the option to not use the Linux driver for operations that can be performed via /dev/mem.

This is exposed to chipsec_utils.py via the command line option "-n".

### DIFF
--- a/source/tool/chipsec/helper/linux/helper.py
+++ b/source/tool/chipsec/helper/linux/helper.py
@@ -78,6 +78,7 @@ IOCTL_MSGBUS_SEND_MESSAGE      = 0x15
 class LinuxHelper(Helper):
 
     DEVICE_NAME = "/dev/chipsec"
+    DEV_MEM = "/dev/mem"
 
     def __init__(self):
         super(LinuxHelper, self).__init__()
@@ -88,6 +89,7 @@ class LinuxHelper(Helper):
         self.os_machine = platform.machine()
         self.os_uname   = platform.uname()
         self.dev_fh = None
+        self.dev_mem = None
 
     def __del__(self):
         try:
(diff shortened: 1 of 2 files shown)


source/tool/chipsec/helper/linux/helper.py
```diff
@@ -137,11 +139,35 @@ def init(self, start_driver):
...
+                                    "Are you running this command as root?\n"
+                                    "%s" % str(err), err.errno)
+        return False
 
     def close(self):
         if self.dev_fh:
             self.dev_fh.close()
         self.dev_fh = None
+        if self.dev_mem:
+            self.dev_mem.close()
```
abazhaniuk: i think should be: os.close(self.dev_mem) instead of self.dev_mem.close()

c7zero:
General comment.

I think it would make sense to implement this consistently for all helper API in generic OSHelper.

Each helper API can thus operate in two modes:
- using CHIPSEC own implementation through kernel module/driver
- using native environment/OS API or capabilities (e.g. through /dev/mem or /sys/firmware/efi etc.) whenever they are available to user mode through OS API or file system.

The mode could be tied to --no_driver command-line argument. By default, all helper API would go through CHIPSEC kernel module. If this option is specified, helper API would use OS native API.

The function is_driver_loaded() in OSHelper would be checked by each helper API function to choose between CHIPSEC and OS native API. I'd rename it to "use_native_mode".

Examples:
- physical memory access via CHIPSEC kernel module vs /dev/mem
- PCIe config access via CHIPSEC kernel module vs /sys/bus/pci/devices
- EFI variable access via CHIPSEC kernel module vs /sys/firmwar/efi/vars on Linux or G/SetFirmwareEnvironmentVariable API on Windows
etc..
- ACPI tables access via CHIPSEC kernel module vs EnumSystemFirmwareTables/GetSystemFirmwareTable on WIndows
etc..

Note that the two modes are also applicable to environments which don't have kernel/user separation, like UEFI shell. These environments may still have their native API which CHIPSEC can use instead its own implementation.

PR #53, opened 2016-08-02
For Linux: load/upload chipsec driver from helper. 

abazhaniuk:
don't need to use  run.sh script. driver will be loaded/unloaded automatically by helper if needed

### DIFF
--- a/source/tool/chipsec/helper/linux/helper.py
+++ b/source/tool/chipsec/helper/linux/helper.py
@@ -113,8 +113,8 @@ class LinuxHelper(Helper):
                     logger().log("Cannot find symbol 'page_is_ram'")
         driver_path = os.path.join(chipsec.file.get_main_dir(), ".." , "drivers" ,"linux", "chipsec.ko" )
         subprocess.check_output( "insmod " + driver_path + " a1=0x%s"%page_is_ram     , shell=True)
-        subprocess.check_output( "chown root:root " + self.DEVICE_NAME            , shell=True)
-        subprocess.check_output( "chmod 600 "       + self.DEVICE_NAME            , shell=True)
+        os.chown(self.DEVICE_NAME, int(os.getenv('SUDO_UID')), int(os.getenv('SUDO_GID')))
+        os.chmod(self.DEVICE_NAME, 600)
         if os.path.exists(self.DEVICE_NAME):
             if logger().VERBOSE:
                 logger().log("Module %s loaded successfully"%self.DEVICE_NAME)
--- a/source/tool/chipsec/helper/linux/helper.py
+++ b/source/tool/chipsec/helper/linux/helper.py
@@ -92,7 +92,6 @@ class LinuxHelper(Helper):
         self.os_machine = platform.machine()
         self.os_uname   = platform.uname()
         self.dev_fh = None
-        self.driver_loaded = False
 
     def __del__(self):
         try:
@@ -120,7 +119,6 @@ class LinuxHelper(Helper):
             os.chown(self.DEVICE_NAME, int(uid), int(gid))
         os.chmod(self.DEVICE_NAME, 600)
         if os.path.exists(self.DEVICE_NAME):
-            self.driver_loaded = True
             if logger().VERBOSE:
                 logger().log("Module %s loaded successfully"%self.DEVICE_NAME)
         else:
--- a/source/tool/chipsec/helper/linux/helper.py
+++ b/source/tool/chipsec/helper/linux/helper.py
@@ -92,6 +92,7 @@ class LinuxHelper(Helper):
         self.os_machine = platform.machine()
         self.os_uname   = platform.uname()
         self.dev_fh = None
+        self.driver_loaded = False
 
     def __del__(self):
         try:
@@ -119,6 +120,7 @@ class LinuxHelper(Helper):
             os.chown(self.DEVICE_NAME, int(uid), int(gid))
         os.chmod(self.DEVICE_NAME, 600)
         if os.path.exists(self.DEVICE_NAME):
+            self.driver_loaded = True
             if logger().VERBOSE:
                 logger().log("Module %s loaded successfully"%self.DEVICE_NAME)
         else:
@@ -140,7 +142,8 @@ class LinuxHelper(Helper):
 
     def stop( self ):
         self.close()
-        subprocess.call(["rmmod", self.MODULE_NAME])
+        if self.driver_loaded:
+            subprocess.call(["rmmod", self.MODULE_NAME])
         if logger().VERBOSE:
             logger().log("[helper] Linux Helper stopped/unloaded")
 
--- a/source/tool/chipsec/helper/linux/helper.py
+++ b/source/tool/chipsec/helper/linux/helper.py
@@ -46,6 +46,8 @@ from chipsec.helper.oshelper import OsHelperError, Helper
 from chipsec.logger import logger, print_buffer
 import errno
 import array
+import subprocess
+import os.path
 import chipsec.file
 
 from ctypes import *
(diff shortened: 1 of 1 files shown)


source/tool/chipsec/helper/linux/helper.py
```diff
@@ -99,7 +103,30 @@ def __del__(self):
...
+    def load_chipsec_module(self):
+        page_is_ram = ""
+        if self.GET_PAGE_IS_RAM:
+            page_is_ram = self.get_page_is_ram()
+            if not page_is_ram:
+                if logger().VERBOSE:
+                    logger().log("Cannot find symbol 'page_is_ram'")
+        driver_path = os.path.join(chipsec.file.get_main_dir(), ".." , "drivers" ,"linux", "chipsec.ko" )
+        subprocess.check_output( "insmod " + driver_path + " a1=0x%s"%page_is_ram     , shell=True)
+        subprocess.check_output( "chown root:root " + self.DEVICE_NAME            , shell=True)
```
tweksteen: Is that necessary? What is the default kernel behaviour? Have you observed older config that behaves differently?

If it is, I suggest using `os.chown` and `os.chmod`.

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -99,7 +103,30 @@ def __del__(self):
...
+    # This function load CHIPSEC driver. (implement functionality from run.sh)
+    def load_chipsec_module(self):
+        page_is_ram = ""
+        if self.GET_PAGE_IS_RAM:
+            page_is_ram = self.get_page_is_ram()
+            if not page_is_ram:
+                if logger().VERBOSE:
+                    logger().log("Cannot find symbol 'page_is_ram'")
+        driver_path = os.path.join(chipsec.file.get_main_dir(), ".." , "drivers" ,"linux", "chipsec.ko" )
+        subprocess.check_output( "insmod " + driver_path + " a1=0x%s"%page_is_ram     , shell=True)
```
tweksteen: You should use a list as first argument to pass the driver path and options. I don't think the shell argument is necessary neither.

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -786,7 +816,12 @@ def get_compression_tool_path( self, compression_type ):
     def getcwd( self ):
         return os.getcwd()
 
-
+    def get_page_is_ram( self ):
+        PROC_KALLSYMS = "/proc/kallsyms"
+        symarr = chipsec.file.read_file(PROC_KALLSYMS).split("\n")
```
tweksteen: use `splitlines()` instead

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -786,7 +816,12 @@ def get_compression_tool_path( self, compression_type ):
...
+        for i in symarr:
```
tweksteen: `s/i/line`

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -786,7 +816,12 @@ def get_compression_tool_path( self, compression_type ):
...
+            if i.find("page_is_ram") != -1:
```
tweksteen: `if "page_is_ram" in line`:

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -99,7 +103,30 @@ def __del__(self):
...
+            if not page_is_ram:
+                if logger().VERBOSE:
+                    logger().log("Cannot find symbol 'page_is_ram'")
+        driver_path = os.path.join(chipsec.file.get_main_dir(), ".." , "drivers" ,"linux", "chipsec.ko" )
+        subprocess.check_output( "insmod " + driver_path + " a1=0x%s"%page_is_ram     , shell=True)
+        os.chown(self.DEVICE_NAME, int(os.getenv('SUDO_UID')), int(os.getenv('SUDO_GID')))
+        os.chmod(self.DEVICE_NAME, 600)
+        if os.path.exists(self.DEVICE_NAME):
+            if logger().VERBOSE:
+                logger().log("Module %s loaded successfully"%self.DEVICE_NAME)
```
tweksteen: Could you please add spaces around the `%` operator?

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -78,6 +80,8 @@
 class LinuxHelper(Helper):
 
     DEVICE_NAME = "/dev/chipsec"
+    MODULE_NAME = "chipsec"
+    GET_PAGE_IS_RAM = True
```
tweksteen: Is this attribute necessary? 
We can always try to find the symbol. If we find it, forward it to the driver. Otherwise, let the driver find it. If the driver fail, report the error and die.
If we want to give the user the choice to decide on that option, we should surface this to chipsec_utils and chipsec_main.

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -109,15 +136,17 @@ def start(self, start_driver):
             logger().log("[helper] Linux Helper started/loaded")
 
     def stop( self ):
+        self.close()
+        subprocess.call(["rmmod", self.MODULE_NAME])
```
tweksteen: Could we use `check_call` instead and log an error if the unloading failed? (We might need to add a condition to test if /dev/chipsec is there before running that. Similarly to what is done in `create`)

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -88,6 +92,7 @@ def __init__(self):
         self.os_machine = platform.machine()
         self.os_uname   = platform.uname()
         self.dev_fh = None
+        self.driver_loaded = False
```
c7zero: parent Helper.__init__ already initializes driver_loaded

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -99,16 +104,46 @@ def __del__(self):
...
+                    logger().log("Cannot find symbol 'page_is_ram'")
+        driver_path = os.path.join(chipsec.file.get_main_dir(), ".." , "drivers" ,"linux", "chipsec.ko" )
+        subprocess.check_output( [ "insmod", driver_path,"a1=0x%s" % page_is_ram ] )
+        uid = os.getenv('SUDO_UID')
+        gid = os.getenv('SUDO_GID')
+        if uid and gid:
+            os.chown(self.DEVICE_NAME, int(uid), int(gid))
+        os.chmod(self.DEVICE_NAME, 600)
+        if os.path.exists(self.DEVICE_NAME):
+            self.driver_loaded = True
```
c7zero: init function sets driver_loaded to True. should probably remove this line from here

source/tool/chipsec/helper/linux/helper.py
tweksteen: Just to clarify on that comment. I was talking about GET_PAGE_IS_RAM. How would you see that GET_PAGE_IS_RAM flag being used?

Should we drop support for kernel older than 2.6.33 ?

source/tool/chipsec/helper/linux/helper.py
```diff
@@ -79,6 +81,8 @@ class LinuxHelper(Helper):
 
     DEVICE_NAME = "/dev/chipsec"
     DEV_MEM = "/dev/mem"
+    MODULE_NAME = "chipsec"
+    GET_PAGE_IS_RAM = True
```
tweksteen: Just to clarify on that comment. I was talking about GET_PAGE_IS_RAM. How would you see that GET_PAGE_IS_RAM flag being used?

Should we drop support for kernel older than 2.6.33 ?

tweksteen:
Just as a side note, it would be good to rebase and squash these commits before merging to master (just to avoid all the "cleanup"-style commits.

PR #59, opened 2016-08-09
Support of native API mode

c7zero:
With this change CHIPSEC can operate in two modes:
1. Using CHIPSEC own kernel driver (default mode, --no_driver option was
not specified)
2. Native OS API mode when helper API can be implemented via native OS
API (--no_driver command-line option was specified)

At the moment, most of the helper API are not implemented in native
mode, except get_ACPI_table in Windows helper.

Helpers, HAL components or modules can use use_native_api() function of
the helper to determine which mode to use. Currently, it directly maps
to --no_driver option. Modules can call this function in is_supported()
if they can only run in either of the modes.

Additional changes:
1. Cleaned up destroy/delete/stop mess in the helpers. Removed destroy()
from the helpers
2. Fixed --exists mode when CHIPSEC works with kernel driver which has
already been loaded


source/tool/chipsec/helper/oshelper.py
```diff
@@ -108,32 +119,39 @@ def loadHelpers(self):
...
+    #
+    # Returns:
+    #   True  if CHIPSEC needs to use native OS/environment API
+    #   False if CHIPSEC needs to use its own (OS agnostic) implementation of helper API
+    #
+    # Currently, CHIPSEC will use native API only if CHIPSEC driver wasn't loaded
+    # (e.g. when --no_driver command-line option is specified).
+    # In future, it can can more conditions
+    # 
+    def use_native_api(self):
```
tweksteen: This is literally the opposite of the `is_driver_loaded` function. Could you remove the `is_driver_loaded` function and replace its calls with `use_native_api`?

tweksteen:
I have slight concern about where the native_mode should be considered in the architecture and I think the ACPI explicitly shows the dilemma: 
 * on Linux, in both cases, we use physical memory to recover the tables either using /dev/chipsec or /dev/mem. In this case, it makes perfect sense that the native_mode is handled by the LinuxHelper.
 * on Windows, we either use physical memory via the driver or the native API of Windows. In this case, the distinction happens with the HAL modules (as shown in your commits).

I am not quite sure how to solve that. But for now, this PR would break Linux ACPI gathering without the driver.

c7zero:
You are right. It turned out to be more complex than I though. Some helper API functions can be made transparently switching between the native and driver modes (like memory access in Linux helper) while other API functions will have to differ on the level of HAL components (like ACPI HAL). In fact, even API that can support both modes may be different. For example, ACPI HAL in driver mode finds ACPI tables directly in memory, so it gets their physical addresses and can extract more tables. ACPI in native mode using Windows GetSystemFirmwareTable would only extract contents of ACPI tables and only the ones Windows exposes to user-mode.

Bottom line, switching between native API mode and chipsec driver mode cannot always be transparently implemented on the helper level (OSHelper or OS specific helpers). HAL will have to be aware of it. I was planning to add use_native_api() to self.cs for HAL to use when needed. Moreover, even to the end consumers of HAL, modules and utils, switching may not always be transparent as well because of aforementioned differences.

For API that can be implemented transparently in the helpers we could switch in OSHelper or in each individual helper. Example of the former case would be something as below: 

    def read_physical_mem( self, phys_address, length ):
        if self.use_native_api():
            return self.helper.native_read_phys_mem(...)
        else:
            return self.helper.read_phys_mem(...)
"native_" version would use /dev/mem or equivalent in other OSes.

The issue is that for all unsupported "native_" functions we'd have to add stubs in all helper. So for now, for the sake of simplicity, I kept the switching in each helper until we have more APIs supporting native mode.

tweksteen:
I agree, the need for `use_native_api` could appear at different location, so having a method in Chipset (`cs`) would help.

I prefer the idea of having the call routed in OSHelper. What you could do is add another condition on the existence of the method:

    def read_physical_mem( self, phys_address, length ):
      if self.use_native_api() and hasattr(self.helper, native_read_phys_mem):
          return self.helper.native_read_phys_mem(...)
      else:
          return self.helper.read_phys_mem(...)

tweksteen:
Looks good to me!

PR #70, opened 2016-08-23
fix path error in chipsec_util in case of installing chipsec to the linux system

abazhaniuk:


source/tool/chipsec_util.py
```diff
@@ -131,8 +131,9 @@ def main(self, argv):
             cmds = map( self.map_modname_zip, filter(self.f_mod_zip, myzip.namelist()) )
         else:
             #traceback.print_stack()
-            mydir = os.path.dirname(__file__)
-            cmds_dir = os.path.join(mydir,os.path.join("chipsec","utilcmd"))
+            import imp
```
tweksteen: Could you move this import at the start of the file?

tweksteen:
Just one small comment. Otherwise, it looks good!

PR #71, opened 2016-08-24
Update chipsec_km.c

thetlk:
```ptr[4] = mdr_out;``` should be outside of braces

See #66 comments


c7zero:
The value of MDR is only returned when transaction expects data back (when direction has MSGBUS_MDR_OUT_MASK). Otherwise, for write only transactions, the value of MDR should not be returned. 0 is a valid value. I'm not sure it would be correct to overwrite whatever value was supplied in ioctl_param with 0 in this case (even though user-mode supplies 0 in msgbus_send_write_message).

c7zero:
Closing this pull request. Please re-open if you think the logic is wrong

PR #72, opened 2016-08-24
Add posibility to free allocated memory

thetlk:
The linux driver doesn't implement memory freeing of allocated memory.
Every allocated memory is saved inside a list. Each time ```IOCTL_FREE_PHYSMEM``` is called, the driver retrieves the virtual address by looking inside the previous allocated memory list. If a valid virtual address if found, the memory buffer is freed. Hence :

- double free is not possible
- only allocated buffer can be freed
- memory leaks are avoided by freeing every buffer still inside the list at the driver cleanup

This simple chipsec module can be used to test this feature :

```python

from chipsec.module_common import *

class test_alloc_free(BaseModule):

	def __init__(self):
		BaseModule.__init__(self)

	def is_supported(self):
		return True

	def run(self, argv):
		l = []
		for i in range(10):
			l.append(self.cs.mem.alloc_physical_mem(0x1000))

		# double free
		print self.cs.mem.free_physical_mem(l[0][1]) # True
		print self.cs.mem.free_physical_mem(l[0][1]) # False

		print self.cs.mem.free_physical_mem(0x1337babe) # False

		for i in range(1, 5):
			print self.cs.mem.free_physical_mem(l[i][1]) # True
```

```
[ 4238.049500] Chipsec module loaded 
[ 4238.049504] ** This module exposes hardware & memory access, **
[ 4238.049505] ** which can effect the secure operation of      **
[ 4238.049506] ** production systems!! Use for research only!   **
[ 4238.142070] [chipsec] freeing pa = 0x20e073000, va = ffff88020e073000
[ 4238.142119] [chipsec] freeing pa = 0x20e076000, va = ffff88020e076000
[ 4238.142134] [chipsec] freeing pa = 0x20e071000, va = ffff88020e071000
[ 4238.142149] [chipsec] freeing pa = 0x20e072000, va = ffff88020e072000
[ 4238.142164] [chipsec] freeing pa = 0x20e077000, va = ffff88020e077000
[ 4238.143760] chipsec cleanup_module 1622: Destroying chipsec device
[ 4238.143826] chipsec cleanup_module 1624: exit
[ 4238.143829] auto freeing allocated memory (va = ffff880222453000, pa = 0x222453000)
[ 4238.143830] auto freeing allocated memory (va = ffff880222451000, pa = 0x222451000)
[ 4238.143831] auto freeing allocated memory (va = ffff880222450000, pa = 0x222450000)
[ 4238.143832] auto freeing allocated memory (va = ffff880222454000, pa = 0x222454000)
[ 4238.143833] auto freeing allocated memory (va = ffff88020e075000, pa = 0x20e075000)
```

### DIFF
--- a/source/drivers/linux/chipsec_km.c
+++ b/source/drivers/linux/chipsec_km.c
@@ -50,6 +50,9 @@ module_param(a1,ulong,0); //a1 is addr of page_is_ram function
 /// Char we show before each debug print
 const char program_name[] = "chipsec";
 
+// list of allocated memory
+struct allocated_mem_list allocated_mem_list;
+
 typedef struct tagCONTEXT {
    unsigned long a;   // rax - 0x00; eax - 0x0
    unsigned long b;   // rbx - 0x08; ebx - 0x4
@@ -1063,7 +1066,9 @@ static long d_ioctl(struct file *file, unsigned int ioctl_num, unsigned long ioc
         //IN params: size
         //OUT params: physical address
         uint32_t NumberOfBytes = 0;
-        void *va, *pa, *max_pa;
+        void *va;
+        phys_addr_t pa, max_pa;
+        struct allocated_mem_list *tmp = NULL;
         
         numargs = 2;
         if(copy_from_user((void*)ptrbuf, (void*)ioctl_param, (sizeof(long) * numargs)) > 0)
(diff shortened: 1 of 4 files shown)


c7zero:
Makes a lot of sense. Thanks!

Issue #13, opened 2015-11-18
Mixed EOL


tweksteen:
Most of the Python source files are using a mixture of LF and CRLF as end-of-line characters. This turns any contribution to the code into an obstacle course. Please pick one or the other and apply to all the source files.

If you decide to go for the CRLF (Windows option), it is as simple as running:

    find . -name \*.py -exec unix2dos {} \;

Once modified, git diff provides the option `--ignore-space-at-eol` to verify that only EOL characters have been adjusted.

c7zero:
Tracking in PR #85

PR #84, opened 2016-09-07
Line endings converted to DOS

andrew-furtak:
Fix for issue #13


abazhaniuk:
For run chipsec_main/chipsec_util from python package in linux line ending should be unix convention.

PR #101, opened 2016-10-11
change setup.py to build driver by default. change root directory of chipsec_main/util

abazhaniuk:


setup.py
```diff
@@ -56,20 +57,22 @@ def package_files(directory):
...
         if platform.system().lower() == "linux" and build_driver:
-            bl = os.path.realpath(self.distribution.command_obj['build'].build_lib)
             subprocess.check_output( "make -C "+ bl +"/drivers/linux/", shell=True )
+        if platform.system().lower() == "linux":
+            shutil.rmtree(bl +"/drivers")
         _build_ext.run(self)
 
-class FullInstall(_install):
-    description = 'Build CHIPSEC Driver. Install CHIPSEC with Driver.'
+class Install_no_driver(_install):
```
tweksteen: Could you please use "install_no_driver" name, without the upper case?

setup.py
```diff
@@ -56,20 +57,22 @@ def package_files(directory):
 class build_ext(_build_ext):
     def run(self):
         global build_driver
+        bl = os.path.realpath(self.distribution.command_obj['build'].build_lib)
         if platform.system().lower() == "linux" and build_driver:
-            bl = os.path.realpath(self.distribution.command_obj['build'].build_lib)
             subprocess.check_output( "make -C "+ bl +"/drivers/linux/", shell=True )
+        if platform.system().lower() == "linux":
+            shutil.rmtree(bl +"/drivers")
```
tweksteen: path.join

setup.py
tweksteen: or CamelCase, but not a mix of both

abazhaniuk:
resubmit new one

PR #109, opened 2016-10-20
Make sure Limit of SPI Protected Range is calculated correctly

timevortex:
"Address bits 11:0 are assumed to be FFFh for the limit comparison" as per reference 21.1.13 on page 730 on http://www.intel.com/content/dam/www/public/us/en/documents/datasheets/8-series-chipset-pch-datasheet.pdf

Therefore I'm ORing 0xFFF to limit, if limit is not 0, to make sure the correct limit is displayed by chipsec.

=================================
Output before this change:
----------------------
[*] BIOS Region: Base = 0x00510000, Limit = 0x00FFFFFF
SPI Protected Ranges
------------------------------------------------------------
PRx (offset) | Value    | Base     | Limit    | WP? | RP?
------------------------------------------------------------
PR0 (74)     | 8FFF0FF0 | 00FF0000 | 00FFF000 | 1   | 0 
PR1 (78)     | 00000000 | 00000000 | 00000000 | 0   | 0 
PR2 (7C)     | 00000000 | 00000000 | 00000000 | 0   | 0 
PR3 (80)     | 00000000 | 00000000 | 00000000 | 0   | 0 
PR4 (84)     | 00000000 | 00000000 | 00000000 | 0   | 0 

=================================
Output with this change:
------------------
[*] BIOS Region: Base = 0x00510000, Limit = 0x00FFFFFF
SPI Protected Ranges
------------------------------------------------------------
PRx (offset) | Value    | Base     | Limit    | WP? | RP?
------------------------------------------------------------
PR0 (74)     | 8FFF0FF0 | 00FF0000 | 00FFFFFF | 1   | 0 
PR1 (78)     | 00000000 | 00000000 | 00000000 | 0   | 0 
PR2 (7C)     | 00000000 | 00000000 | 00000000 | 0   | 0 
PR3 (80)     | 00000000 | 00000000 | 00000000 | 0   | 0 
PR4 (84)     | 00000000 | 00000000 | 00000000 | 0   | 0

PR #114, opened 2016-10-26
SPI HAL cleanup to use XML config

c7zero:
Removed hardcoded offsets from hal.spi when reading FREGx and PRx and
replaced with XML config.

This pull request also includes PR #109 (Make sure Limit of SPI Protected
Range is calculated correctly) by @timevortex except bits 11:0 of PRx
limit are also set to FFFh when PRL field is 0 (in this case PRx applies
to the first 4kB of flash).


timevortex:
I've created PR #116 to fix a minor change. You should be able to easily merge that into your spi_hal_cleanup branch before merging that into master.

PR #115, opened 2016-10-26
Improve build_ext

tweksteen:
We have been relying a lot on package_data to include data when building the Python package. This is fine for some data (i.e., .cfg, .ini, .json) because these may appear at different location within the chipsec module. It is not fine to use package_data for any other data, especially data residing outside the chipsec subfolder. This is by design of how that attribute works. (For now, that attribute has been abused by using an extra parent reference ".." in the value. This literally relies on the parent directory of the project which makes little sense).

Another avenue to include to include extra data is to use the data_files attribute. This should be used for static data that may or may not be part of a package. The perfect example for that attribute is the chipsec-manual.pdf file at the root of the directory. What is interesting is that files described in this section are not copied when building the package but only when installing (which make sense as we expect them to be static).

The driver(s) source code does not fall in any of the two previous category and we therefore need another way to include/declare it for compilation at build time. For that purpose, I've added some code within the existing build_ext class to handle the copy of the source to the build directory and triggering the Makefile there.


c7zero:
I'm good with changes to setup.py and removing run.sh is fine too.

We need to be careful with changes to driver and cores.so Makefiles. CHIPSEC is also being used as a standalone utility. So we need a simple way of building the driver and cores.so "in place" without installing is as a package via setup.py

I plan to add scripts/build.sh script which will build chipsec.ko, copy it to helper location and build cores.so. This way we can remove unnecessary instructions from drivers/linux/Makefile and have a script to build chipsec in place.

PR #121, opened 2016-11-14
Add a check for primary to sideband bridge hide.

hcwang1126:
Primary to sideband bridge device is to be hidden by the BIOS before PCI enumeration step.
After post, the unhiding should be done within SMM only to prevent OS from seeing it.


johnloucaides:
After discussing with a few folks, we don't believe hiding or not hiding P2SB constitutes a vulnerability. Software can, and in some cases should, hide/unhide this as part of normal functionality. Instead of a test module, we're planning to actually enable the use of this functionality. You can see the start of this in accepted PR #200 .

Please have a look and reply whether or not you agree with this path. Thanks!

c7zero:
Agree with @johnloucaides 
Closing PR. Please re-open if think it should be a separate module

PR #124, opened 2016-11-17
Use HALBase for all HAL modules

tweksteen:
Clean up the dependencies between HAL modules and chipset.py. The previous imports were hiding circular dependencies because most of the HAL modules were relying on the global `chipset` object. By using a parent class, all the reference to `chipsec.chipset` are removed in favour of `self.cs`.

Fix util commands and modules accordingly.

A large chunk of module methods from chipset.py and mmio.py are moved to object methods (i.e., fix the indentation). It might be useful to review this PR by ignoring the space changes (i.e., `git diff -w HEAD^` or  adding `?w=1` to the github pull request URL)

### DIFF
--- a/chipsec/hal/iommu.py
+++ b/chipsec/hal/iommu.py
@@ -28,9 +28,7 @@ Access to IOMMU engines
 
 from chipsec.logger import *
 
-import chipsec.hal.iobar
-import chipsec.hal.mmio
-import chipsec.hal.paging
+from chipsec.hal import hal_base, mmio, paging
 
 IOMMU_ENGINE_DEFAULT = 'VTD'
 IOMMU_ENGINE_GFX     = 'GFXVTD'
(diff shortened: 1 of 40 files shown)


chipsec/hal/iommu.py
```diff
@@ -45,32 +43,33 @@
...
     pass
 
-class iommu:
-    def __init__( self, cs ):
-        self.cs = cs
+class IOMMU(hal_base.HALBase):
+
+    def __init__(self, cs):
+        super(IOMMU, self).__init__(cs)
+        self.mmio = mmio.MMIO(cs)
```
c7zero: This should probably be removed from iommu class and replaced with self.cs.mmio

chipsec/hal/spi.py
```diff
@@ -142,65 +142,67 @@
...
+class SpiRuntimeError (RuntimeError):
+    pass
+class SpiAccessError (RuntimeError):
+    pass
+
+class SPI(hal_base.HALBase):
+
+    def __init__(self, cs):
+        super(SPI, self).__init__(cs)
+        self.mmio = mmio.MMIO(cs)
```
c7zero: should use self.cs.mmio instead of self.mmio

chipsec/utilcmd/mmcfg_cmd.py
```diff
@@ -52,10 +52,11 @@ def requires_driver(self):
 
     def run(self):
         t = time.time()
+        _mmio = mmio.MMIO(self.cs)
 
         if 2 == len(self.argv):
-            #pciexbar = get_PCIEXBAR_base_address( self.cs )
-            pciexbar = get_MMCFG_base_address( self.cs )
+            #pciexbar = _mmio.get_PCIEXBAR_base_address()
+            pciexbar = _mmio.get_MMCFG_base_address()
```
c7zero: Same: s/_mmio/self.cs.mmio

chipsec/utilcmd/mmio_cmd.py
```diff
@@ -72,16 +73,16 @@ def run(self):
         t = time.time()
 
         if ( 'list' == op ):
-            list_MMIO_BARs( self.cs )
+            _mmio.list_MMIO_BARs()
```
c7zero: s/_mmio/self.cs.mmio

chipsec/modules/tools/cpu/sinkhole.py
```diff
@@ -38,17 +38,17 @@ class sinkhole(BaseModule):
 
     def __init__(self):
         BaseModule.__init__(self)
-        self._cpu = chipsec.hal.cpu.CPU( self.cs )
+        self._cpu = cpu.CPU(self.cs)
```
c7zero: should remove this and use self.cs.cpu insteand of self._cpu

c7zero:
Accepting regardless of self.cs.mmio and self.cs.cpu comments. This is long overdue improvement.
Thanks @tweksteen !

PR #130, opened 2016-12-02
fix bug when calls compression tools on linux

abazhaniuk:


chipsec/helper/oshelper.py
```diff
@@ -466,7 +466,7 @@ def decompress_file( self, CompressedFileName, OutputFileName, CompressionType )
           exe = self.get_compression_tool_path( CompressionType )
           if exe is None: return None 
           try:
-            subprocess.call( '%s -d -o %s %s' % (exe,OutputFileName,CompressedFileName), stdout=open(os.devnull, 'wb') )
+            subprocess.call( ('%s -d -o %s %s' % (exe,OutputFileName,CompressedFileName)).split(" "), stdout=open(os.devnull, 'wb') )
```
tweksteen: You don't need to go through the list representation here. Something like the following would be cleaner:

    subprocess.call([exe, "-d", "-o", OutputFileName, CompressedFileName], stdout=...

PR #132, opened 2016-12-05
add DKMS support

abazhaniuk:
added support DKMS:
- added dkms.conf file with configuration for DKMS (without module installation) 
- updated linux helper. First it checks default location of the driver: chipsec/helper/linux/chipsec.ko if driver is not there helper tries to find it in DKMS modules storage. if driver there helper uses it from that location, if not then helper didn't find driver and informs about it by logger().error.

### DIFF
--- a/chipsec/helper/linux/helper.py
+++ b/chipsec/helper/linux/helper.py
@@ -107,12 +107,6 @@ class LinuxHelper(Helper):
 ###############################################################################################
 # Driver/service management functions
 ###############################################################################################
-    def get_currect_kernel_version(self):
-        return subprocess.check_output("uname -r", shell=True)
-
-    def get_architecture(self):
-        return subprocess.check_output("uname -m", shell=True)
-
     def get_version(self):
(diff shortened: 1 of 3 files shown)


chipsec/helper/linux/helper.py
```diff
@@ -119,6 +125,11 @@ def load_chipsec_module(self):
             else:
                 a1 = "a1=0x%s" % page_is_ram 
         driver_path = os.path.join(chipsec.file.get_main_dir(), "chipsec", "helper" ,"linux", "chipsec.ko" )
+        if not os.path.exists(driver_path):
+            #check DKMS modules location
+            driver_path = self.get_dkms_module_location()
+            if not os.path.exists(driver_path):
+                logger().error("Cannot find chipsec.ko module")
```
tweksteen: Should it raise an exception here on top of logging the error?

c7zero:
I'm ok

tweksteen:
Looks good to me!

PR #133, opened 2016-12-06
make chipsec_tools executable in Linux

abazhaniuk:
setup.py makes LzmaCompress.bin and TianoCompress.bin binaries executable during installation.


tweksteen:
This should already be the case (working fine on my local install) : http://stackoverflow.com/questions/18409296/package-data-files-with-executable-permissions

Could you be more specific about the use case that is failing for you?

abazhaniuk:
To make chipsec easier to use in case if someone copy chipsec from windows to linux,.

c7zero:
Any concerns with merging this PR?
The issue is in a very specific case when chipsec is installed on Linux from a folder created under Windows (e.g. chipsec was downloaded on Windows), UEFI parsing functionality breaks because compression tools aren't executable. So one would have to manually make tools executable. After this change, setup.py would make all necessary tools executable.

tweksteen:
This sounds like a really particular case that makes little sense to handle in Chipsec. Could you get the source directly on Linux? Could you copy the archive (.zip) to the Linux machine without extracting it?

PR #134, opened 2016-12-13
UEFI firmware parsing improved

c7zero:
- Creating EFI module object hierarchy when parsing firmware images
- Exporting firmware object hierarchy and results of modules (--json option) to JSON format
- Improved EFI modules search


tweksteen:
It would be great to start adding some unit test on the UEFI parsing part of the code. You could create a simplistic rom image and call `chipsec_util uefi decode` on it, similarly to what is done with other utilcmd. This would ensure that large refactoring like this one don't break anything.

c7zero:
Makes sense. The challenge has been that UEFI parsing can break on certain images so having one reference image may not do any good. We should see what's the best way to implement a unit test for this functionality. Perhaps having a few images from various vendors. This PR has been tested on ~80 full or update images from various vendors so it should be good for now.

PR #144, opened 2017-01-25
Added missing Atom C2000 Processor Family Device IDs

ErikBjorge:
I was not able to test these new IDs as I do not have any hardware.  The IDs are from the Atom C2000 Processor Family Microserver Datasheet.


c7zero:
Table 10-6 of Intel Atom Processor C2000 Product Family for Microserver Datasheet (https://www-ssl.intel.com/content/dam/www/public/us/en/documents/datasheets/atom-c2000-microserver-datasheet.pdf)

PR #152, opened 2017-02-07
Ported efi_compressor from uefi-firmware-parser to chipsec_tools

abazhaniuk:


setup.py
```diff
@@ -145,6 +145,9 @@ def initialize_options(self):
 install_requires = []
 extra_kw = {}
 
+compression_source_files = []
```
tweksteen: Since these variables are only used in the 'elif', could you move their initialization into that block?

tweksteen:
One small comment, otherwise it looks good to me.

c7zero:
Nice change!
Should implement compress/decompress_buffer and compress_file to OShelper in future in addition to decompress_file.

CC @theopolis

PR #153, opened 2017-02-08
Added missing map_io_space method in UEFI helper code.

ErikBjorge:
Added the missing map_io_space method to resolve common.bios_wp
failures when executing under the UEFI Shell.


c7zero:
Not sure if we need separate pa2va helper API (none of other helpers have pa2va) but otherwise have no  problems.. Nice catch.

PR #156, opened 2017-02-23
IGD hal component added. igddma utility command added.

andrew-furtak:
Internal graphics device component added to hal. Utility command implemented to read/write memory using graphics device DMA.

### DIFF
--- a/chipsec/cfg/common.xml
+++ b/chipsec/cfg/common.xml
@@ -171,7 +171,7 @@
     <register name="PCI0.0.0_CAPID0_B"   type="pcicfg" bus="0" dev="0" fun="0" offset="0xE8" size="4" desc="Capabilities B"/>
 
     <!-- IGD registers -->
-    <register name="IGD_DID2" type="pcicfg" bus="0" dev="2" fun="0" offset="2" size="2" desc="Device Identification Number">
+    <register name="PCI0.2.0_DID" type="pcicfg" bus="0" dev="2" fun="0" offset="2" size="2" desc="Device Identification Number">
     </register>
 
     <!-- LPC Interface Bridge -->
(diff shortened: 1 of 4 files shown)


chipsec/cfg/common.xml
```diff
@@ -170,6 +170,10 @@
     <register name="PCI0.0.0_CAPID0_A"   type="pcicfg" bus="0" dev="0" fun="0" offset="0xE4" size="4" desc="Capabilities A"/>
     <register name="PCI0.0.0_CAPID0_B"   type="pcicfg" bus="0" dev="0" fun="0" offset="0xE8" size="4" desc="Capabilities B"/>
 
+    <!-- IGD registers -->
+    <register name="IGD_DID2" type="pcicfg" bus="0" dev="2" fun="0" offset="2" size="2" desc="Device Identification Number">
```
c7zero: Can you please rename IGD DID register name to "PCI0.2.0_DID" to be consistent with other devices

chipsec/hal/igd.py
```diff
@@ -0,0 +1,201 @@
...
+        return (pa & ~0xFFF)
+
+    def get_PA_from_PTE(self, pte):
+        if self.is_legacy_gen():
+            return self.get_PA_from_PTE_legacy()
+        else:
+            return self.get_PA_from_PTE_gen8( pte )
+
+
+    def gfx_aperture_dma_read_write(self, address, size=0x4, value=None, pte_num=0):
```
c7zero: Would probably be better to swap arguments 'value' and 'pte_num' so that a caller could do read through arbitrary PTE without specifying named 'pte_num' argument.

chipsec/hal/igd.py
```diff
@@ -0,0 +1,201 @@
...
+        super(IGD, self).__init__(cs)
+        self.helper = cs.helper
+        self.is_legacy = None
+        self.enabled = None
+
+    def __identify_device(self):
+        if self.enabled is None:
+            self.is_legacy = False
+            try:
+                self.dev_id = self.cs.pci.read_word( 0, 2, 0, 2 )
```
c7zero: Should use cs.read_register('PCI0.2.0_DID') instead

chipsec/hal/igd.py
```diff
@@ -0,0 +1,201 @@
...
+#
+# CHIPSEC: Platform Hardware Security Assessment Framework
+#
+# -------------------------------------------------------------------------------
+
+"""
+Working with Intel processor Integrated Graphics Device (IGD)
+
+usage:
+    >>> gfx_aperture_dma_read(0x80000000, 0x100)
```
c7zero: Should fix usage documenting both read and write cases via gfx_aperture_dma_read_write

PR #181, opened 2017-03-16
Fix --no_driver option on macOS

c7zero:


tweksteen:
Great! Looks good to me!

PR #182, opened 2017-03-16
Fixed exception when running tools.smm.smm_ptr in UEFI Shell.

ErikBjorge:
Exception was caused by efi.allocphysmem returning a tuple but the
helper code was expecting a numeric value.  Updated UEFI helper code to
select the correct value in the tuple.


c7zero:
Accepting, as the fix would not require recompilation of Python for EFI shell. Thanks!
But, I believe, this should properly be fixed by replacing [this line](https://github.com/chipsec/chipsec/blob/master/chipsec_tools/edk2/PythonEFI/edk2module.c#L6962) with
`return PyLong_FromUnsignedLong((unsigned long)va);`

PR #196, opened 2017-04-14
PFS support and parsing fixes.

andrew-furtak:


abazhaniuk:
this conflict with #192 because of: , errors="ignore" - this part should be added as part of fix of #192

PR #200, opened 2017-04-19
Add MMIO register access for P2SB.

ErikBjorge:


c7zero:
One concern with the implementation is that it prescribes specific mechanism to send messages on the message bus through primary-to-sideband bridge if SBREGBAR register is defined in XML. So there's no way to use MCR/MDR based mechanism to access the message bus.

In other words, it tells consumers how they *should* access specific HW, not how they *can* access it. The intent is to have the latter. CHIPSEC consumers/modules may want to explicitly send messages via MCR/MDR or may want to use both mechanisms and compare the results. For example, this is why CHIPSEC allows access to PCIe configuration registers through both memory-mapped CFG and legacy CF8/CFC.
  
I'd suggest to add a flag in msgbus_reg_read/write functions which would allow callers to choose between the two mechanisms. I'd also add new commands (e.g. mread/mwrite) to "chipsec_util msgbus"

ErikBjorge:
Thanks for the feedback on this.  I will update the code with your suggestions and supply a patch next week.

ErikBjorge:
After looking at the code again I think this is how I would like to implement the update.

- Create a new register type mm_msgbus.  This allows the platform to select how to access the register in the XML and matches the PCIe mmcfg register distinction.
- Create new functions mm_msgbus_reg_read and mm_msgbus_reg_write and move the new functionality to these functions.  I would like to do this because the mm_msgbus port width is a byte not a word.
- Add mread/mwrite commands to chipsec_util msgbus.

We may also need to look at the register definitions for msgbus because I think some fields have moved between BYT and 100/200 series chipsets.

Please let me know if you think this is reasonable.

PR #206, opened 2017-04-22
Updated README and added Erik Bjorge to AUTHORS

c7zero:


c7zero:
We probably don't need README

PR #213, opened 2017-05-04
Fixed detection of successful SMBUS transactions.

ErikBjorge:
A successful transaction is detected when the busy bit transitions from 1 to 0 and the intr bit is set.  If the busy bit transitions from a 1 to 0 without setting the intr bit an error has occurred.


ErikBjorge:
Yes, it was not correctly detecting failed transactions when an SPD was not present.  On a failed transaction it was incorrectly returning the contents of the data register even though the transaction failed.  In most cases this was fine because the last byte of a valid SPD tended to be 0xFF and appeared to be an error case.  By checking the INTR bit we know that the transaction completed successfully.

Here is the description of the INTR bit from the 200 series PCH EDS.
Interrupt: (INTR): When set, this indicates that the source of the interrupt or SMI# was the successful completion of its last command.

This bit is set regardless if interrupts/SMI# is enabled for the device.

Tested on Broadwell and Skylake platforms.

PR #217, opened 2017-05-09
Add register util command

tweksteen:
Add a register util command to read the value of a specific register. Only the read operation is implemented at this stage.


c7zero:
Awesome! I think print_register should be routed through "chipsec_util reg print"

PR #220, opened 2017-05-09
Updated code to load common*.xml files first.

ErikBjorge:
Also enforced a consistent (alphabetical) load order.

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -371,17 +371,39 @@ class Chipset:
     ##################################################################################
 
     def init_xml_configuration( self ):
+        # Create a sorted config file list (xml only)
+        _cfg_files = []
         _cfg_path = os.path.join( chipsec.file.get_main_dir(), 'chipsec/cfg' )
-        # Load chipsec/cfg/common.xml configuration XML file common for all platforms if it exists
-        self.init_cfg_xml( os.path.join(_cfg_path,'common.xml'), self.code )
-        # Load chipsec/cfg/<code>.xml configuration XML file if it exists for platform <code>
+        for root, subdirs, files in os.walk(_cfg_path):
+            _cfg_files.extend([os.path.join(root, x) for x in files if fnmatch.fnmatch(x, '*.xml')])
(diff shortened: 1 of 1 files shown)


ErikBjorge:
Any comments or concerns @c7zero that you can see with this patch?

chipsec/chipset.py
```diff
@@ -371,17 +371,40 @@ def use_native_api(self):
...
+                    self.init_cfg_xml(_xml, self.code)
+
         # Load configuration from all other XML files recursively (if any)
-        for dirname, subdirs, xml_fnames in os.walk( _cfg_path ):
-            for _xml in xml_fnames:
-                if fnmatch.fnmatch( _xml, '*.xml' ) and not fnmatch.fnmatch( _xml, 'common.xml' ) and not (_xml in ['%s.xml' % c.lower() for c in Chipset_Code]):
-                    self.init_cfg_xml( os.path.join(dirname,_xml), self.code )
+        if logger().VERBOSE: logger().log("[*] Loading Other Configurations:")
+        for plat in [c.lower() for c in Chipset_Code]:
+            loaded_files.extend([x for x in _cfg_files if fnmatch.fnmatch(os.path.basename(x), '{}*.xml'.format(plat))])
```
c7zero: It seems to me that it's better to not extend loaded_files here with <plat>*.xml files which weren't really loaded. Maybe use a separate list exclude_files in addition to loaded_files? The reason is that later we may want to list all loaded XML files and we would use loaded_files list for that and it would contain XML files which weren't really loaded.

ErikBjorge: Sounds good.  I think I will generate the loaded_file list first and then load all of the files after it is generated.  Will update this change today.

ErikBjorge:
Correct, the load order you described is what this fix will produce.

ErikBjorge:
New changes for you @c7zero to look at.  Let me know if you want any log information along with this change for clarification.

frinzell:
Updates look good.

c7zero:
Looks good to me.

Issue #204, opened 2017-04-20
helper/driver interface thread_id size


johnloucaides:
We are using 1 byte to hold the number of threads when we pack structs between python and windows driver. This breaks on systems with >255 threads. For example:
[    def read_msr( self, cpu_thread_id, msr_addr ):](https://github.com/chipsec/chipsec/blob/master/chipsec/helper/win/win32helper.py#L634)
        (eax,edx) = (0,0)
        out_length = 8
        out_buf = (c_char * out_length)()
        in_buf = struct.pack( '=BI', cpu_thread_id, msr_addr )
        out_buf = self._ioctl( IOCTL_RDMSR, in_buf, out_length )
        (eax, edx) = struct.unpack( '2I', out_buf )
        return (eax, edx)

I think we need to just use a larger size. I didn't notice this much in linux (mostly seems to be 4 byte value), but we need to check more thoroughly. Threads are not supported by chipsec in uefi shell right now.

ErikBjorge:
Closing this issue with fix PR #230.  Please reopen if the problem was not resolved.

PR #228, opened 2017-06-05
Fix for issue #204

BrentHoltsclaw:


ErikBjorge:
Why is this done in three commits?  Would you rather squash them into a single commit?

BrentHoltsclaw:
User error, yes should be only one commit

PR #238, opened 2017-06-10
More operation in chipsec_util reg command

c7zero:
Added write, read_field, write_field, get_control, set_control operations.
Makes it easier to work with registers by names

### DIFF
--- a/tests/software/test_reg.py
+++ b/tests/software/test_reg.py
@@ -33,5 +33,13 @@ class TestRegChipsecUtil(util.TestChipsecUtil):
         self._chipsec_util("reg read PCI0.0.0_TOUUD LOCK")
         self._assertLogValue("PCI0.0.0_TOUUD.LOCK", "0x0")
 
+    def test_reg_read_field1(self):
+        self._chipsec_util("reg read_field PCI0.0.0_TOUUD LOCK")
+        self._assertLogValue("PCI0.0.0_TOUUD.LOCK", "0x0")
+
+    def test_reg_get_control(self):
+        self._chipsec_util("reg get_control FlashLockDown")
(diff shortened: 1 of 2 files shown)


tests/software/test_reg.py
```diff
@@ -37,9 +37,9 @@ def test_reg_read_field1(self):
         self._chipsec_util("reg read_field PCI0.0.0_TOUUD LOCK")
         self._assertLogValue("PCI0.0.0_TOUUD.LOCK", "0x0")
 
-    def test_reg_get_control(self):
-        self._chipsec_util("reg get_control FlashLockDown")
-        self._assertLogValue("FlashLockDown", "0x0")
+    #def test_reg_get_control(self):
+    #    self._chipsec_util("reg get_control FlashLockDown")
```
tweksteen: The second argument of `_chipsec_util` is a MockHelper that only implements the HAL functions we are expecting to call. If it is not supplied, the standard [TestHelper](https://github.com/chipsec/chipsec/blob/master/tests/software/mock_helper.py#L51) will be used which only defines PCI read on 0:0:0. For the `FlashLockDown` register, we can use the default [SPIHelper](https://github.com/chipsec/chipsec/blob/master/tests/software/mock_helper.py#L219).

In this case, that test can easily be fixed by changing this line to:

        self._chipsec_util("reg get_control FlashLockDown",
                           mock_helper.SPIHelper)

ErikBjorge:
Based on previous conversations @c7zero it sounded like you were going to remove the support to read a register field as part of the 'read' operation.  If someone wants to read a field they can use the read_field operation.  Is this still the plan?

c7zero:
Yes but let's file a new PR

PR #247, opened 2017-06-21
Merge FREGx definitions

tweksteen:
With 2aa81af, the definitions of flash registers started to be duplicated. When using a live dump (`chipsec_util.py spi dump`), the `common.xml` definitions were still used. On SkyLake and KabyLake, the incorrect size of FREG.RB was then loaded (13 bits instead of 15).

This commit unifies the definitions. That is, FREG0_FLASHD is replaced by FLREG0. This does not affect `spidesc` as only the fields are used and not the bar attribute of the registers.


tweksteen:
Thanks for the explanation, this makes sense. 

I will close this PR for now and create a new one in which I'll add the FREG* definitions for SKL/KBL. This should address the initial issue.

PR #250, opened 2017-06-29
Add mmap functionality to kernel module

deja-nmooney:


deja-nmooney:
Looks like the Travis build didn't run, but seems to be due to an issue with apt rather than the changes I'd like to merge

ErikBjorge:
We will get the code reviewed and then force another build.

deja-nmooney:
Hi all,
just curious if I could get an update on the status of this PR. Thanks!

ErikBjorge:
Can @c7zero @abazhaniuk or @johnloucaides take a look at this change and provide any feedback?

johnloucaides:
Are you just trying to enable mmap in the linux helper or did you want to add it to util so that it works from command line as well?

Also, I started reading up a bit on phys_mem_access_prot... should we be using this all over other accesses or just with mmap?

deja-nmooney:
This is just adding mmap support to the actual `/dev/chipsec` device, so that it can be mmaped like other files. `/dev/mem` supports this functionality, and it's very useful when you want to map a section of physical memory into virtual memory. I am working on a project that requires arbitrary physical memory R/W + access to some of the other ChipSec IOCTLs, so this seems like the best way to go about it. The alternative is seeking/reading/writing, but the mmap approach is much cleaner.

The only changes I made to the helper are to support finding the `phys_mem_access_prot` symbol, which is not exported by the kernel (but is used by the `/dev/mem` device -- `char.c` in the Linux source). Usually this is done via kallsyms lookup, but the address of the function can also be manually passed to the module, and the ChipSec helper will support this functionality if this pull request is merged. The functionality of the helper/util otherwise stays the same.

My understanding is that `phys_mem_access_prot` is only used to set flags on a new virtual memory mapping, so it is unnecessary for the rest of ChipSec's functionality. You can look at the [Linux mem.c source](https://github.com/torvalds/linux/blob/master/drivers/char/mem.c) to see how `phys_mem_access_prot` is used in implementing `/dev/mem`.

PR #253, opened 2017-07-05
Fix the range check in bios_wp.py

fukai-t:
This patch is a bug fix.

bios_wp.py checked protection ranges if they protect the BIOS range or not.
However, when a protection range was same as the BIOS range, the result was wrong.


fukai-t:
Thank you for your response.

> Btw, have you seen a system that actually hit this case?

No, I haven't.
I tried to change the protect range by my code running in the kernel mode.

PR #256, opened 2017-07-19
Rwe fixups

johnloucaides:
Currently, if you use the RWE driver, it's a little annoying to remember to set an environment variable. Instead, we can just look to see which driver is actually present underneath the helper directory. For RWE we also look in the system drivers directory so that if RWE is installed, the driver can be found in the default location.

Additionally, if you try to use decode or whitelist, the RWE helper is missing decompress methods. This just copies them from the regular windows helper.

### DIFF
--- a/chipsec/helper/rwe/__init__.py
+++ b/chipsec/helper/rwe/__init__.py
@@ -20,7 +20,8 @@
 #
 
 import platform, os
-if "windows" == platform.system().lower() and (os.path.isfile(os.path.join("C:\\Windows\\System32\\drivers\\RwDrv.sys")) or os.path.isfile(os.path.join(os.getcwd(), "win7_amd64\RwDrv.sys"))):
+
+if "windows" == platform.system().lower():
     __all__ = [ "rwehelper" ]
 else:
     __all__ = [ ]
--- a/chipsec/helper/rwe/__init__.py
+++ b/chipsec/helper/rwe/__init__.py
@@ -21,8 +21,6 @@
 
 import platform, os
 if "windows" == platform.system().lower() and (os.path.isfile(os.path.join("C:\\Windows\\System32\\drivers\\RwDrv.sys")) or os.path.isfile(os.path.join(os.getcwd(), "win7_amd64\RwDrv.sys"))):
-    print "rwe"
     __all__ = [ "rwehelper" ]
 else:
-    print "no rwe"
     __all__ = [ ]
--- a/chipsec/helper/rwe/__init__.py
+++ b/chipsec/helper/rwe/__init__.py
@@ -20,7 +20,9 @@
 #
 
 import platform, os
-if "windows" == platform.system().lower() and os.environ.has_key("USERWE"):
+if "windows" == platform.system().lower() and (os.path.isfile(os.path.join("C:\\Windows\\System32\\drivers\\RwDrv.sys")) or os.path.isfile(os.path.join(os.getcwd(), "win7_amd64\RwDrv.sys"))):
+    print "rwe"
     __all__ = [ "rwehelper" ]
 else:
+    print "no rwe"
     __all__ = [ ]
--- a/chipsec/helper/rwe/rwehelper.py
+++ b/chipsec/helper/rwe/rwehelper.py
@@ -72,10 +72,11 @@ kernel32 = windll.kernel32
 
 drv_hndl_error_msg = "Cannot open rwe driver handle. Make sure rwe driver is installed and started if you are using option -e (see README)"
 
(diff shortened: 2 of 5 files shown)


chipsec/helper/win/win32helper.py
```diff
@@ -353,6 +353,8 @@ def create(self, start_driver):
         if logger().VERBOSE: logger().log( "[helper] service control manager opened (handle = 0x%08x)" % hscm )
 
         driver_path = os.path.join( chipsec.file.get_main_dir(), "chipsec", "helper", "win", self.win_ver, DRIVER_FILE_NAME )
+        if not os.path.isfile( driver_path ):
+            driver_path = os.path.join("C:\\Windows\\System32\\drivers\\RwDrv.sys"
```
c7zero: I think it's a bad idea to add interdependence between the two helpers. I'd rather windows helper fail if it doesn't find required driver than attempting to load a driver from another helper.

chipsec/helper/rwe/__init__.py
```diff
@@ -20,7 +20,7 @@
 #
 
 import platform, os
-if "windows" == platform.system().lower() and os.environ.has_key("USERWE"):
+if "windows" == platform.system().lower() and (os.path.isfile(os.path.join("C:\\Windows\\System32\\drivers\\RwDrv.sys")) or os.path.isfile(os.path.join(os.getcwd(), "win7_amd64\RwDrv.sys"))):
```
c7zero: Why look in system drivers directory at all? It never is there.

c7zero: and agree that looking for USERWE var is annoying ;)

c7zero: Also, I'm not sure it will work when chipsec is built with py2exe. The sys file is not with the rwehelper package. Should probably use "os.path.join( chipsec.file.get_main_dir(), "chipsec", "helper", "rwe", self.win_ver, DRIVER_FILE_NAME )" instead just like rwehelper.py

johnloucaides: driver should be in the system32/drivers directory if rwe is installed through the normal installer... at least, that's what happened in my system and why I added it in the first place. 

Agree on the join for the path of the chipsec location. I'll commit a fix soon.

chipsec/helper/win/win32helper.py
johnloucaides: Hum. Good point. I think the idea was to allow someone to put the chipsec driver in system32/drivers as an alternative to in the chipsec directory. Would you rather see a commit that removes this or one that fixes it to look for the chipsec driver under system32?

johnloucaides:
The latest commit should clean up some of the above. It seems cleaner to move the logic for determining what to do if you have both into the helpers.py file. If you do have both, it will prefer the chipsec driver. Both RWE and chipsec helpers will look in system32/drivers as well as chipsec helper subdir.

Please have another look when you get a chance. Thanks!

chipsec/helper/rwe/rwehelper.py
```diff
@@ -72,10 +72,11 @@ def cfg_address(self):
 
 drv_hndl_error_msg = "Cannot open rwe driver handle. Make sure rwe driver is installed and started if you are using option -e (see README)"
 
-DRIVER_FILE_NAME = "RwDrv.sys"
-DEVICE_FILE      = "\\\\.\\RwDrv"
-SERVICE_NAME     = "RwDrv"
-DISPLAY_NAME     = "RwDrv"
+DRIVER_FILE_PATHS   = ["C:\\Windows\\System32\\drivers", os.path.join( chipsec.file.get_main_dir(), "chipsec", "helper", "rwe", "win7_" + platform.machine().lower())]
```
ErikBjorge: It might be a good idea to use os.path.join for the fixed system path as well.

chipsec/helper/win/win32helper.py
```diff
@@ -73,6 +73,7 @@ class PCI_BDF(Structure):
 
 drv_hndl_error_msg = "Cannot open chipsec driver handle. Make sure chipsec driver is installed and started if you are using option -e (see README)"
 
+DRIVER_FILE_PATHS = [ "C:\\Windows\\System32\\drivers", os.path.join( chipsec.file.get_main_dir(), "chipsec", "helper", "win", "win7_" + platform.machine().lower()) ]
```
ErikBjorge: Same comment about using os.path.join for the fixed path.

ErikBjorge:
Tested the change and is seems to work just fine.  Any issues with merging this update?

PR #273, opened 2017-08-22
Added a function "search" to look for a signature in memory. Ex: chip…

hcwang1126:
it's convenient for a BIOS engineer to dump some resource tables.

### DIFF
--- a/chipsec/utilcmd/mem_cmd.py
+++ b/chipsec/utilcmd/mem_cmd.py
@@ -132,13 +132,14 @@ class MemCommand(BaseCommand):
         elif 'search' == op and len(self.argv) > 5:
             phys_address = int(self.argv[3],16)
             size         = int(self.argv[4],16)
-            target = 0
-            for i in range(0, len(self.argv[5])):
-                target += ord(self.argv[5][i]) << (i * 8)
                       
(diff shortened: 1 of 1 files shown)


chipsec/utilcmd/mem_cmd.py
```diff
@@ -128,6 +129,17 @@ def run(self):
...
 
+        elif 'search' == op and len(self.argv) > 5:
+            phys_address = int(self.argv[3],16)
+            size         = int(self.argv[4],16)
+            target = 0
+            for i in range(0, len(self.argv[5])):
+                target += ord(self.argv[5][i]) << (i * 8)
+                      
+            buffer = self.cs.mem.read_physical_mem( phys_address, size )
+            offset = buffer.find(self.argv[5])
```
ErikBjorge: Should handle the case where the search string is not found and -1 is returned.

ErikBjorge:
Other than the one small change I think this works well.  Used it to manually find the ACPI anchor and table signatures.

chipsec/utilcmd/mem_cmd.py
hcwang1126: Agree.

hcwang1126:
Already commit a fix to handle that the signature is not found.

PR #284, opened 2017-09-09
Migrate driver project to Visual Studio and closes #283

### LINKED ISSUE #283: IOCTL_WRITE_PHYSMEM/IOCTL_WRITE_PHYSMEM don't parse the physical address properly.
read_phys_mem() and write_phys_mem() pack the structure as the following, which means the lowPart of the physical address is gonna be first and not second. Just like the way PHYSICAL_ADDRESS structure is ordered.
```
in_buf = struct.pack( '<4IQ', phys_address_lo, phys_address_hi, len(buf), 0, addressof(write_buf) )
```
e.g.:
```
>>> in_buf = pack( '<4IQ', 0x41414141, 0x42424242, 0x43434343, 0, 666)
>>> in_buf
'AAAABBBBCCCC\x00\x00\x00\x00\x9a\x02\x00\x00\x00\x00\x00\x00'
```

And the Windows driver side parse the input buffer, by assuming the first DWORD is the `highPart` of the physical address instead of the `lowPart`.
```
            pInBuf = Irp->AssociatedIrp.SystemBuffer;
            pOutBuf = Irp->AssociatedIrp.SystemBuffer;

            phys_addr.HighPart = ((UINT32*)pInBuf)[0];
            phys_addr.LowPart  = ((UINT32*)pInBuf)[1];
            len                = ((UINT32*)pInBuf)[2];
```

msuiche:
- It's 2017, so MAKEFILEs are from the past.
And closes the parsing issues with #283.


johnloucaides:
I don't have an issue with moving to VS in principle, but want to make sure we're not messing with people's workflows. Could the assigned reviewers weigh in?

ErikBjorge:
I don't mind enabling VS2017 support but I do have a few other comments.
* This patch adds VS2017 support as well as makes changes to the driver.  This should be two patches.  For now just submit the support files for VS2017.
* We still need to support WinDDK build because it is a freely available tool set.
* We need to verify if this project can be built using the community version of VS2017 and does not require a professional version to build.

BrentHoltsclaw:
I agree with @ErikBjorge .

ErikBjorge:
I would still like to see two changes instead of one.  The switch to VS2017 is fine.  The changes in driver.c and chipsec.h should be its own change.  Would it be OK @abazhaniuk and @msuiche if we submit two new changes and then close this pull request?

abazhaniuk:
Good for me.

abazhaniuk:
Any progress on this PR?

ErikBjorge:
I would rather see this as two changes.

BrentHoltsclaw:
Closing PR driver updated in PR#462

PR #290, opened 2017-09-21
Expose TPM commands and state to chipsec_utils

BrentHoltsclaw:


tweksteen:
Could you:
- Fix the code so that the current unit tests passes
- Add a unit test for at least one subcommand
- Change the commit message to something more useful (like "Expose TPM commands and state to chipsec_utils....")
Otherwise, this looks good!

tweksteen:
My bad here, the unit test expects `parse_eventlog` command (https://github.com/chipsec/chipsec/blob/master/tests/software/test_tpm_eventlog.py#L37). Could you please fix this test as well?

chipsec/utilcmd/tpm_cmd.py
```diff
@@ -11,27 +12,65 @@
...
+        
+        Examples:
+                        
+        >>> chipsec_util tpm parse_log binary_bios_measurements
+        >>> chipsec_util tpm state 0
+        >>> chipsec_util tpm command pcrread 0 17
+        >>> chipsec_util tpm command continueselftest 0
+        """
+        def requires_driver(self):
+            return False
```
ErikBjorge: Need to load driver in some cases (command and state) but not in others.

chipsec/utilcmd/tpm_cmd.py
```diff
@@ -11,27 +12,65 @@
...
 
+        def run(self):
```
ErikBjorge: Need to create an instance of the TPM HAL class.  See the SPI command as an example.

_tpm = TPM(self.cs)

chipsec/utilcmd/tpm_cmd.py
```diff
@@ -11,27 +12,65 @@
...
+                tpm_eventlog.parse(log)
+            elif ('command' == op ):
+                if len(self.argv) < 5:
+                    print TPMCommand.__doc__
+                    return
+                tpm.command( self.argv[3], self.argv[4], self.argv[5:] )
+            elif ('state' == op ):
+                tpm.dump_access ( self.argv[3] )
+                tpm.dump_status ( self.argv[3] )
+                tpm.didvid ( self.argv[3] )
```
ErikBjorge: Should be dump_didvid

chipsec/utilcmd/tpm_cmd.py
```diff
@@ -11,27 +12,65 @@
...
+#along with this program; if not, write to the Free Software
+#Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+#Contact information:
+#chipsec@intel.com
+#
 
 from chipsec.command import BaseCommand
 from chipsec.hal import tpm_eventlog
+from chipsec.hal import tpm
```
ErikBjorge: TPM class in HAL should be derived from hal_base.HALBase.

chipsec/utilcmd/tpm_cmd.py
```diff
@@ -11,27 +12,77 @@
...
-            print TPMCommand.__doc__
-            return
-        log = open(self.argv[3])
-        tpm_eventlog.parse(log)
+        def run(self):
+            try:
+                _tpm = tpm.TPM(self.cs)
+            except TpmRuntimeError, msg:
+                print(msg)
+                return
```
ErikBjorge: It looks like we have some extra indentation here...  Same with the chunk above.  I will pull the code to check.

Issue #269, opened 2017-08-18
decompression weird behavior


abazhaniuk:
when decode 3440a02.rom check:
[3440a02.zip](https://github.com/chipsec/chipsec/files/1235509/3440a02.zip)

in 3440a02.rom.dir/1_180000-7FFFFF_BIOS.bin.dir/FV/01_8C8CE578-8A3D-4F1C-9935-896185C32DD3.dir/146_7D113AA9-6280-48C6-BACE-DFE7668E8307.FV_FREEFORM.dir/ 
After decompression we have:
3440a02.rom.dir/1_180000-7FFFFF_BIOS.bin.dir/FV/01_8C8CE578-8A3D-4F1C-9935-896185C32DD3.dir/146_7D113AA9-6280-48C6-BACE-DFE7668E8307.FV_FREEFORM.dir/00_S_COMPRESSION.dir/00_S_UNKNOWN_97 
But when i run again i got different file:
3440a02.rom.dir/1_180000-7FFFFF_BIOS.bin.dir/FV/01_8C8CE578-8A3D-4F1C-9935-896185C32DD3.dir/146_7D113AA9-6280-48C6-BACE-DFE7668E8307.FV_FREEFORM.dir/00_S_COMPRESSION.dir/00_S_UNKNOWN_8D 
Content a bit different as well. 
when i try to decompress this file by hand i have different file (consistent file)

abazhaniuk:
if we take a look at: 
https://chromium.googlesource.com/chromiumos/platform/vboot_reference/+/0.12.433.B62/utility/efidecompress.c

We can see: 

  // Try new version first
  r = TianoDecompress(ibuf, isize, obuf, osize, sbuf, ssize);
  if (r != EFI_SUCCESS) {}
  // Try old version
  r = EfiDecompress(ibuf, isize, obuf, osize, sbuf, ssize);
  if (r != EFI_SUCCESS) {} 

first check if TianoDecompress can decompress and then EfiDecompress, so in our case we need to fix: 

 elif CompressionType == 0x01:
            data = self.decompress_data( [ EFI, Tiano ], CompressedFileData )

to:

 elif CompressionType == 0x01:
            data = self.decompress_data( [ Tiano, EFI ], CompressedFileData )

but it will case crashes in library on other UEFI images, like 
[P11-B2.zip](https://github.com/chipsec/chipsec/files/1252866/P11-B2.zip)

abazhaniuk:
[...]

PR #294, opened 2017-10-06
Fix decompression issue #269

PerezFederico:
Fix for the issue and some other small things.

Basically, as Alex said, when you swap the decompression algorithms order:

_chipsec/helper/linux/helper.py_

    data = self.decompress_data( [ Tiano, EFI ], CompressedFileData )

You have to check for overflow to check if the algorithm is correct for that file and then return and error code to inform that the algorithm was correct:

_chipsec_tools/compression/Tiano/Decompress.c_

    if (DataIdx > Sd->mOrigSize) {
        Sd->mBadAlgorithm = 1;
        return;
        }

Then:

    if (Sd->mBadTableFlag != 0 || Sd->mBadAlgorithm != 0) {
    //
    // Something wrong with the source
    //
        Status = EFI_INVALID_PARAMETER;
   }

Until here it is as Alex proposed. The hidden thing was that there is also an overflow in the mPTTable buffer in the Scratch struct (when the algorithm is wrong) which was stepping into the malloc header of the Destination buffer. This was causing a crash when you attempted to free the *Destination bacause the header was corrupt.

    Status = Extract((VOID *)SrcBuf, SrcDataSize, (VOID **)&DstBuf, &DstDataSize, type);
    if (Status != EFI_SUCCESS) {
        PyErr_SetString(PyExc_Exception, "Failed to decompress\n");
        errorHandling(SrcBuf, DstBuf);    //<--- CRASH
        return NULL;
   }
 So, if we expand the mPTTable buffer to withstand this wrong data filling, we have the problem solved.

The other changes are pretty self explanatory, if you have questions just ask.
Fede.

### DIFF
--- a/chipsec/helper/linux/helper.py
+++ b/chipsec/helper/linux/helper.py
@@ -107,6 +107,9 @@ class LinuxHelper(Helper):
     SUPPORT_KERNEL26_GET_PHYS_MEM_ACCESS_PROT = False
     DKMS_DIR = "/var/lib/dkms/"
 
+    decompression_oder_type1 = [Tiano, EFI]
+    decompression_oder_type2 = [LZMA, Tiano, EFI]
+
     def __init__(self):
         super(LinuxHelper, self).__init__()
         self.os_system  = platform.system()
@@ -1034,13 +1037,26 @@ class LinuxHelper(Helper):
             if "phys_mem_access_prot" in line:
                return line.split(" ")[0]
 
-    def decompress_data(self, funcs, cdata):
+    def rotate_list(self, list, n):
+        return list[n:] + list[:n]
+
+    def rotate_algorithms(self, CompressionType, n):
+        if(CompressionType == 0x01):
+            self.decompression_oder_type1 = self.rotate_list(self.decompression_oder_type1, n)
+        elif(CompressionType == 0x02):
+            self.decompression_oder_type2 = self.rotate_list(self.decompression_oder_type2, n)
+
+    def decompress_data(self, funcs, cdata, CompressionType):
+        failed_times = 0
         for func in funcs:
             try:
                 data = func(cdata, len(cdata))
-                return  data
+                if(failed_times > 0):
+                    self.rotate_algorithms(CompressionType, failed_times)
+                return data
             except Exception:
-                continue
+                failed_times += 1
+                continue      
         return None
     #
     # Decompress binary with efi_compressor from https://github.com/theopolis/uefi-firmware-parser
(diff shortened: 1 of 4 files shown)


PerezFederico:
Sorry for all the commits, they were just for me, didn't think they were going to appear all here.

tweksteen:
Could you use git rebase (with the squash option) and repush to your branch? Thanks

PerezFederico:
:)

abazhaniuk:
Looks good, need to test it more. Let me run tests over weekend.

BrentHoltsclaw:
I don't understand how increasing the size fixes the issue.  Increasing the buffer size for mPTTable buffer may allow the crash file to successfully complete, but it seems like it may only mask the issue.  Could this still fail with other inputs?

PerezFederico:
I think it's a fixed overflow if the data is wrong. But let me check. If it's not the case I will have to put the corresponding checks for the boundaries. I did not wanted to do that because of the overhead.

PerezFederico:
Yes, you made me think I can't be sure the overflow will not be bigger. I added the checks now, and restored the buffer to the original size. Now, the check I did for the overflow in the last commit it's kind of redundant because I'm checking it in an early stage and giving the same error. Let me know what you think.

PerezFederico:
-  if(Index > TableSize)
    + if(Index >= TableSize)

So I was stepping 1 byte when we stimulate with wrong data but it seems it did not make any errors but now it is correct.

abazhaniuk:
Run test with ~15k bios images. let me analyze result of it tomorrow. if everything good - I will merge it tomorrow.

abazhaniuk:
Question. Do we want to leave brute force method when we running decompression, like:
data = self.decompress_data( [ Tiano, EFI ], CompressedFileData )
or we want to detect which of the algo is using and use it. 
Other idea, what if we check two decompression methods just for first binary and save this result and for result BIOS images (with type=1) we use just this algo. so we detect which one in first binary and apply this result to result of binaries.

PerezFederico:
Second suggestion is clear and I think it is good. I will do that and also change what BrentHoltsclaw suggest about TableBits.
For first part, how do you suggest I should detect which algorithm should I use? Tiano and EFI decompression use the same algorithm, the only difference is the mPbit value, that it is the number of bits to encode the offset. The only way I can think right now is to do as you said, if one algorithm fails by the method of the overflow, just use the other method for the rest of the binaries. The only problem should be if we receive a binary with the other compression method once we already detected a wrong one. Maybe we can switch methods every time an algorithm fails so we make sure it always works.

PerezFederico:
Implemented as discussed. Every time an algorithm fails, we rotate to the next algorithm on the list, and stay like that until we fail again.

For example:
 
For file: https://github.com/chipsec/chipsec/files/1235509/3440a02.zip
The default order  [Tiano, EFI] is ok. 

For file: https://github.com/chipsec/chipsec/files/1252866/P11-B2.zip
We switch algorithms near the end.

Neither case we see a significant performance improvement, but in cases when EFI is the required algorithm we should better results. I don't have the files to test it extensively.

The other changes are because as suggested 1<<TableBits give the table size. I did not noticed that before, so now it is better.

Issue #310, opened 2017-10-25
TypeError: decompress_data() takes exactly 4 arguments (3 given)


tweksteen:
Running `chipsec_util decode image.bin` raises (tested on Chipsec 1.3.4):

      [...] linux/helper.py", line 1071, in decompress_file
      data = self.decompress_data( self.decompression_oder_type2, CompressedFileData )
    TypeError: decompress_data() takes exactly 4 arguments (3 given)

This works fine on version 1.3.3.

PerezFederico:
I added this bug. I will fix it.

ErikBjorge:
This missing parameter should be fixed in v1.3.5.

PR #312, opened 2017-10-26
Fix decompression bug

PerezFederico:
Fix a decompression issue causing crashes and probably a security vulnerability.
It is a fix also for something I did wrong in another commit that was merged.


PerezFederico:
I also fixed Issue #310 that was added by myself. Sorry.

BrentHoltsclaw:
Line 781 should handle this case (where compression size is larger than the input buffer), however it does not account for the fact that compsize +8 could overflow and cause fillbuf to read outside of the input like you are seeing.

PerezFederico:
Let me see if I understand.
You mean this?

       if (SrcSize < CompSize + 8) {
         return EFI_INVALID_PARAMETER;
         }

To this?

       if (SrcSize <= CompSize + 8) {
         return EFI_INVALID_PARAMETER;
         }

BrentHoltsclaw:
I am proposing adding a check for compsize + 8 to ensure it did not overflow (i.e. 0xFFFFFFF9 +8).  Something like this:
```
if (SrcSize < CompSize + 8 || (CompSize + 8) < 8 ) {
     return EFI_INVALID_PARAMETER;
     }
```

PerezFederico:
I understood now. Done. Good catch.

ErikBjorge:
So for some reason it seems that VERSION was included on the branch and when merged it lost our development tag.  This is not a big deal but could have been worse if it happened with some other code changes.  In the future it might be better to rebase the branch before merging.  I tend to use the rebase and merge option to detect this type of conflict.  Rebase and merge also forces a linear history.

No worries, I have submitted a new pull request to fix this very minor issue.

PR #342, opened 2018-02-07
Change Linux READ/WRITE Physical Memory Functionality

BrentHoltsclaw:


drivers/linux/chipsec_km.c
```diff
@@ -1254,6 +1254,67 @@ static long d_ioctl(struct file *file, unsigned int ioctl_num, unsigned long ioc
...
         break;
     }
+
+	case IOCTL_READ_PHYSMEM:
+	{
+		// IN params : physical address, length
+		// OUT params : buffer
+		printk( KERN_INFO "[chipsec] > READMEM\n");
+
+		unint64_t NumberofBytes = 0;
```
ErikBjorge: Did you mean uint64_t in this case.  The CI build is breaking.

drivers/linux/chipsec_km.c
```diff
@@ -1254,6 +1254,67 @@ static long d_ioctl(struct file *file, unsigned int ioctl_num, unsigned long ioc
...
+		break;
+	}
+
+	case IOCTL_WRITE_PHYSMEM:
+	{
+		// IN params : physical address, length, ptr to buffer
+		// OUT params : length
+		printk( KERN_INFO "[chipsec] > WRITEMEM\n");
+
+		unint64_t NumberofBytes = 0;
```
ErikBjorge: Same thing here with the uint64_t.

PR #349, opened 2018-02-16
Add module to check HW debug features on/off

abazhaniuk:


chipsec/modules/debugenabled.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+    def __init__(self):
+        BaseModule.__init__(self)
+
+    def is_supported(self):
+        current_platform_id = self.cs.get_chipset_id()
+        supported = (current_platform_id == chipsec.chipset.CHIPSET_ID_CFL) | (current_platform_id == chipsec.chipset.CHIPSET_ID_KBL) | (current_platform_id == chipsec.chipset.CHIPSET_ID_SKL)
+        if not supported: self.logger.log_skipped_check( "DCI is not supported in this platform" )
+        return supported
```
ErikBjorge: You are skipping the entire test if DCI is not supported.  Would you mind changing the is_supported function to just verify that the IA32_DEBUG_INTERFACE MSR exists and then check the platform before calling check_dci.  This gives us some better test coverage on older systems.

You can use CPU ID to determine if the IA32_DEBUG_INTERFACE MSR is supported on the platform using (CPUID (1,0) checking ecx bit 11).

chipsec/modules/debugenabled.py
```diff
@@ -0,0 +1,120 @@
...
+        dci_test_fail = self.check_dci();
+        cpu_debug_test_fail = self.check_cpu_debug_enable();
+            
+        if dci_test_fail == True:
+            self.logger.log_failed_check( 'DCI Debug is enabled' )
+            returned_result = ModuleResult.FAILED
+        if cpu_debug_test_fail == True:
+            self.logger.log_failed_check( 'CPU IA32_DEBUG_INTERFACE is enabled' )
+            returned_result = ModuleResult.FAILED
+
```
ErikBjorge: The log_failed_check should only be called once in a test.  Can you change this to log_failed for each of the blocks and then add a new entry below that calls log_failed_checked with a string like 'One or more debug interfaces enabled.' if returned_result is set to ModuelResult.FAILED.

abazhaniuk:
fixed. check out.

ErikBjorge:
Thanks for the updates.  Looks good to me but want a review from @johnloucaides.

johnloucaides:
I recall some false positives with debug_occurred... I noticed this on SKL NUC. The 2 other SKL machines I tried report that the feature is not present in cpuid. Maybe more testing could explain. In the meantime, I recommend not returning failure on debug_occurred alone. Options may be to just have an informational print or possibly return warning.

PR #355, opened 2018-02-22
Fixed error reporting when warnings are found.

ErikBjorge:
Discovered that the warning path was not working as expected. Showed
failed in log message but warning in summary.


ErikBjorge:
Made some small changes to make sure the functions return a consistent type to fix the error reporting issue.  Also only log with status (PASSED/WARNING/FAILED) at the end of the test.  I also found some of the verbose messages useful on a normal test run so made them visible.

Please retest @abazhaniuk on some of your target systems and let me know if anything needs to be changed.

PR #357, opened 2018-02-23
Fixed build warnings in driver.

ErikBjorge:


ErikBjorge:
Reverting driver change so this update will not be needed at this time.

PR #359, opened 2018-02-23
fix exception when reading physical memory

johnloucaides:
Previously, testing chipsec_util to read physical memory would give the following error:
  File ".../chipsec/chipsec/helper/linux/helper.py", line 371, in read_phys_mem
    ret = struct.unpack(str(length)+'s', outstr)
struct.error: unpack requires a string argument of length 16


ErikBjorge:
We may still have some issues but I need to do some more testing.  Not sure if this is related to the python code or the driver.  Will continue to look at the problem over the weekend.

Test:
1. Allocate memory using chipsec_util
1. Read buffer multiple times using chipsec_util and verify consistent values
1. Write dword into buffer using chipsec_util
1. Read buffer multiple times using chipsec_util and verify consistent values

for step 2 I got consistent values but in step 4 I did not and then my system crashed.

PR #369, opened 2018-03-14
Add ability to only display result deltas from previous test run.

ErikBjorge:
This allow the user to check for changes in test results or over time.
This can include the additon or removal of test as well as changes in
test results.

### DIFF
--- a/chipsec/result_deltas.py
+++ b/chipsec/result_deltas.py
@@ -1,0 +1,76 @@
+#CHIPSEC: Platform Security Assessment Framework
+#Copyright (c) 2018, Intel Corporation
+#
+#This program is free software; you can redistribute it and/or
+#modify it under the terms of the GNU General Public License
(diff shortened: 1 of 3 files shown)


tweksteen:
How does this work if I just want to compare two JSON results without re-running the test suite? I would also suggest extracting the comparison code and store it in a separate file/method within chipsec/.

ErikBjorge:
Thanks for the feedback @tweksteen. I will split the code out into its own module.  Would using chipsec_util be fine for just checking two JSON log files?

ismaws:
How about presenting/exporting results in XML format?

ErikBjorge:
I can look at adding XML output support as well.  I do not want to deal with using XML as an input format for log file deltas.

tweksteen:
Yep, that sounds good to me. Thanks for implementing this!

Issue #363, opened 2018-03-01
Kernel module does not load in Fedora running kernel v4.15.4-300 and above


ettermyra:
When loading the CHIPSEC kernel module in Fedora kernel 4.15.4-300 and above the error "Unknown symbol in module" is produced. This applies to at least version 1.3.6RC1 and 1.3.5 of CHIPSEC.

When loading in Fedora kernel 4.14.16 it all works just fine! Unfortunately I do not know what has changed from one kernel version to another...Some patches related to Meltdown and Spectre?...just specualting since I have not read the release notes for the kernel.

johnloucaides:
I'll try this when I get to my Fedora system, but can you copy/paste which symbol is "unknown"?

ettermyra:
Oh! Sorry about that! =P Here goes (dmesg):
Unknown symbol valid_mmap_phys_addr_range (err 0)

petersandor:
I'm experiencing the same issue on Arch 4.15.12-1

ErikBjorge:
Thanks for the information @petersandor.  I am going to have @BrentHoltsclaw take a look at it.

BrentHoltsclaw:
I created a pull request Fix for issue #363.  If you have time, could you check to see if it fixes your issue.

ettermyra:
Tried the suggested fix and initial results indicate that it works fine. Will do some more testing...

johnloucaides:
Are we ok to get this merged? Kinda annoying to have Fedora broken. ;-)

ErikBjorge:
It was merged on the 17th.

johnloucaides:
My bad. It wasn't this error that I noticed. It was that Fedora 27 now needs *another* dependency (elfutils-libelf-devel). Sorry.

I've updated the linux install instructions here: https://github.com/chipsec/chipsec/wiki/Installing-CHIPSEC-in-Linux

ettermyra:
Sorry for the delay!
All seem to be working just fine! Closing this issue.
Thanks!

PR #371, opened 2018-04-03
Fix for issue #363

BrentHoltsclaw:
Defined function valid_mmap_phys_addr_range within linux driver if
kernel version is greater than 4.15.4

### DIFF
--- a/drivers/linux/chipsec_km.c
+++ b/drivers/linux/chipsec_km.c
@@ -544,6 +544,13 @@ int __weak phys_mem_access_prot_allowed(struct file *file,
         return 1;
 }
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,15,4)
+int valid_mmap_phys_addr_range(unsigned long pfn, size_t size)
+{
+	return 1;
+}
+#endif
+
 #ifndef ARCH_HAS_VALID_PHYS_ADDR_RANGE
 static inline int valid_phys_addr_range(phys_addr_t addr, size_t count)
 {


drivers/linux/chipsec_km.c
```diff
@@ -544,6 +544,13 @@ int __weak phys_mem_access_prot_allowed(struct file *file,
         return 1;
 }
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,15,4)
```
ErikBjorge: Do we need to worry about the case where ARCH_HAS_VALID_PHYS_ADDR_RANGE is not defined?  If so the function will be defined twice.

drivers/linux/chipsec_km.c
```diff
@@ -544,6 +544,13 @@ int __weak phys_mem_access_prot_allowed(struct file *file,
...
+int valid_mmap_phys_addr_range(unsigned long pfn, size_t size)
```
ErikBjorge: Function signature is different than the original, is this on purpose?

BrentHoltsclaw: asm/io.h defines the function as follows:
#define ARCH_HAS_VALID_PHYS_ADDR_RANGE
extern u64 kern_mem_attribute (unsigned long phys_addr, unsigned long size);
extern int valid_phys_addr_range (phys_addr_t addr, size_t count); /* efi.c */
extern int valid_mmap_phys_addr_range (unsigned long pfn, size_t count);

When defined as static inline there were compilation errors due to mismatched function types.

drivers/linux/chipsec_km.c
BrentHoltsclaw: Agreed.  Would #undef ARCH_HAS_VALID_PHYS_ADDR_RANGE be a better solution?

drivers/linux/chipsec_km.c
ErikBjorge: Thanks for checking that.

drivers/linux/chipsec_km.c
ErikBjorge: Yes, update the change to check the kernel version and that ARCH_HAS_VALID_PHYS_ADDR_RANGE is defined before adding your function definition.

ErikBjorge: @BrentHoltsclaw can you do some testing on a few different distros to make sure this does not break anything.  If it looks good we can go with this solution for now and fix any corner cases if they show up.

BrentHoltsclaw:
@ErikBjorge was able to load the driver within ARCH, and Fedora.  Lets push and handle corner cases if needed.

PR #391, opened 2018-05-18
Create sgx_check.py

dcg-red-team:
SGX configuration check module


dcg-red-team:
Module to check SGX feature and configuration

PR #392, opened 2018-05-18
linux driver physmem write fix

BrentHoltsclaw:
copy_from_user was only zeroing bytes in some cases when physmem write
function was called.  Changed to a for loop which appears to fix the
issue.


frinzell:
Almost forgot.  Copyright date.  :)

PR #393, opened 2018-05-18
Module to check SGX configuration and capabilities

dcg-red-team:


chipsec/cfg/common.xml
```diff
@@ -590,6 +590,39 @@
     <register name="MSR_LT_LOCK_MEMORY" type="msr" msr="0x2E7" desc="LT lock memory configuration MSR">
       <field name="LT_LOCK"  bit="0" size="1" desc="Lock bit" />
     </register>
+    <register name="BIOS_SE_SVN" type="msr" msr="0x302" desc="BIOS SGX SVN values">
```
ErikBjorge: Can you add a comment to this block of registers to indicate the CPUID function to call before accessing the registers?  It could be helpful for future test writers.

dcg-red-team:
Thanks for the feedback.  Included comments in common.xml file. to check if CPUID(EAX=7,ECX=0):EBX[2] is set before accessing SGX registers

PR #400, opened 2018-06-07
Updated processing of variables with cert/key information

ErikBjorge:
Not all cert/key variables have the same format.  Added individual processing routines for each variable format.  Additionally updated missing GUIDs for cert/key/hash types.


frinzell:
Only caught that the copyright year needs updating.  Thanks!

PR #405, opened 2018-06-15
Update hide_p2sb routine

frinzell:
Hidden test is more universal and register write less destructive.


ErikBjorge:
@BrentHoltsclaw can you review this change?

PR #419, opened 2018-07-18
General cleanup

BrentHoltsclaw:
Print and exception cleanup


chipsec/helper/oshelper.py
```diff
@@ -32,6 +32,7 @@
 Abstracts support for various OS/environments, wrapper around platform specific code that invokes kernel driver
 """
 
+from six import with_metaclass
```
frinzell: When running this, I get error message: no module named six

BrentHoltsclaw:
Breaking into smaller commits

PR #420, opened 2018-07-19
Fix for uefi_fuzzer force crash

Michaelm396IN:
Originally, the uefi_fuzzer would crash due to the overallocation of resources, specifically in response to the objects _NAME & _DATA. In response, I created a bounds method which limits the size of any information being passed. I also decreased the amount of generated data with the use of modulo (tested for with limit value in mind). The test now runs without failure.

### DIFF
--- a/chipsec/modules/tools/uefi/uefivar_fuzz.py
+++ b/chipsec/modules/tools/uefi/uefivar_fuzz.py
@@ -188,10 +188,8 @@ class uefivar_fuzz(BaseModule):
                     _DATA = None
                     while not _DATA:
                         _DATA   = self.rnd(random.randrange(BOUND_INT))
-                print _DATA
                         
                 if FUZZ_SIZE  : _SIZE   = random.randrange(len(_DATA))
-                print ("hoe",_SIZE)            
                 
                 if (count < CASE): continue
                 
--- a/chipsec/modules/tools/uefi/uefivar_fuzz.py
+++ b/chipsec/modules/tools/uefi/uefivar_fuzz.py
@@ -105,8 +105,7 @@ class uefivar_fuzz(BaseModule):
         self.logger.log( USAGE_TEXT )
         return True
 
-    def bound(self,value): 
-        BOUNDS = 1000
+    def bound(self,value,BOUNDS):
         if value:
             if sys.getsizeof(value) < BOUNDS:
                 return value
@@ -131,6 +130,7 @@ class uefivar_fuzz(BaseModule):
         ITERATIONS = 1000
         SEED       = int(time())
         CASE       = 1
+        BOUND      = 1000 #tested value that can be increased or decreased to fit the limit bounds
 
 
         FUZZ_NAME   = True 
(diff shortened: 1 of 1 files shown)


chipsec/modules/tools/uefi/uefivar_fuzz.py
```diff
@@ -103,14 +103,15 @@ def rnd(self, n=1):
     def usage(self):
         self.logger.log( USAGE_TEXT )
         return True
-
+
```
ErikBjorge: Added some trailing white space.

chipsec/modules/tools/uefi/uefivar_fuzz.py
```diff
@@ -169,25 +172,33 @@ def run( self, module_argv ):
...
-                if FUZZ_DATA  : _DATA   = self.rnd(int(self.rnd(3),16))
-                if FUZZ_SIZE  : _SIZE   = int(self.rnd(3),16)
-
+                
+                if FUZZ_DATA  : 
+                    _DATA = None
+                    while not _DATA:
+                        _DATA   = self.rnd(random.randrange(BOUND_INT))
+                        
+                if FUZZ_SIZE  : _SIZE   = int(self.rnd(3),16)
```
ErikBjorge: Do we want the buffer length to be <= to the actual buffer size?  We know that the runtime interface has no way to validate that the buffer size matches the actual buffer.

chipsec/modules/tools/uefi/uefivar_fuzz.py
```diff
@@ -160,6 +162,7 @@ def run( self, module_argv ):
                 else: help_text = self.usage()
         
         if not help_text:
+
```
ErikBjorge: Remove extra blank line

chipsec/modules/tools/uefi/uefivar_fuzz.py
```diff
@@ -169,25 +172,33 @@ def run( self, module_argv ):
...
             self.logger.log('')
-     
             for count in range(1,ITERATIONS+CASE):
                 
                 if FUZZ_NAME:
                     _NAME = ''
-                    for n in range(int(self.rnd(3),16)):
-                        _NAME += random.choice(string.printable)
+                    while not _NAME:
+                        for n in xrange(random.randrange(BOUND_STR)):
```
frinzell: Recommend using range() for future Python 3 compatibility.

chipsec/modules/tools/uefi/uefivar_fuzz.py
Michaelm396IN: Are you suggesting we change the FUZZ_SIZE to be eqaul to _SIZE = len(_DATA)?

ErikBjorge: We could set _SIZE = random.randrange(len(_DATA)) so that they may not match but we will not overflow our python buffer.

chipsec/modules/tools/uefi/uefivar_fuzz.py
```diff
@@ -169,25 +172,35 @@ def run( self, module_argv ):
...
+                
+                if FUZZ_DATA  : 
+                    _DATA = None
+                    while not _DATA:
+                        _DATA   = self.rnd(random.randrange(BOUND_INT))
+                print _DATA
+                        
+                if FUZZ_SIZE  : _SIZE   = random.randrange(len(_DATA))
+                print ("hoe",_SIZE)            
+
```
ErikBjorge: Remove extra prints of _DATA and _SIZE

PR #421, opened 2018-07-19
Python logger integration

Michaelm396IN:
In this update, I re-factored the original Logging method for python's built in logging module (imported as "pythonlogging"). The new module also aids in FileHandleing to any desired output file and StreamHandeling to the console. The resulting data is seamlessly parallel to the original logging method but now uses an intuitive logging practice to log information to the console or file. Additionally, color logging is integrated using the module "coloredlogs". This should eliminate the need to specify windows or linux system with the integration of coloredlogs.install() (not yet tested on windows).

### DIFF
--- a/chipsec/logger.py
+++ b/chipsec/logger.py
@@ -307,10 +307,9 @@ class Logger:
     def log( self, text):
         """Sends plain text to logging."""
         #self.rootLogger.info(text)
-        print ("{}".format(text))
-        #self._log(text, None, None)
+        self._log(text, None, None)
 
-    """
+    
     def _log(self, text, color, isStatus):
         "Internal method for logging"
         if self.LOG_TO_FILE: self._save_to_log_file( text )
@@ -321,7 +320,7 @@ class Logger:
                 if self.ALWAYS_FLUSH: sys.stdout.flush()
         if self.xmlAux.useXML: self.xmlAux.append_stdout(text)
         #if isStatus: self._save_to_status_log_file( text )
-    """
+    
 
     def error( self, text ):
         """Logs an Error message"""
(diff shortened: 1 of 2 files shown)


chipsec/modules/common/memlock.py
```diff
@@ -51,7 +51,7 @@ def is_supported(self):
         # causes a GP fault and can't be caught in UEFI Shell.
         if self.cs.get_chipset_id() in chipsec.chipset.CHIPSET_FAMILY_ATOM:
             return False
-        return True
+        return False
```
ErikBjorge: Looks like you may have committed an extra file.  This should not be added to this commit.

chipsec/logger.py
```diff
@@ -30,7 +30,8 @@
 """
 Logging functions
 """
-
+import logging as pyLogging
+import coloredlogs
```
ErikBjorge: I have a feeling this is causing the CI build problem.  You may want to just submit the logging changes and leave the color coding alone for now.  We can then look at updating the color logging.

chipsec/logger.py
```diff
@@ -87,13 +88,13 @@ def log_color( fg_color, text ):
             """
             old_setting = WConio.gettextinfo()[4] & 0x00FF
             WConio.textattr( COLOR_ID[ fg_color ] )
-            print text
+            print("{}".format(text))
```
ErikBjorge: print(text) would work as well.

chipsec/logger.py
```diff
@@ -391,15 +382,14 @@ def start_test( self, test_name ):
...
+        self.rootLogger.critical(text)
         self.xmlAux.start_test( test_name )
 
 
     def start_module( self, module_name ):
         """Displays a banner for the module name provided."""
-        #text = "\n[*] start module: %s" % module_name
-        #self._log(text, WHITE, None)
-        self.log( "\n[*] running module: %s" % module_name )
+        text = "\n[*] running module: %s" % module_name
```
ErikBjorge: Convert to using format

PR #433, opened 2018-08-30
New method of defining pcicfg registers - 'device' argument tells whi…

astepkowski:
…ch pci device shoulkd be used to determine exact location of register


ErikBjorge:
We will update the XSD file in another pull request.

PR #436, opened 2018-09-11
Change % to .format

BrentHoltsclaw:
Initial pass at changing % to .format.  
  - Missed a few such as where the join function is used.  
  - Still working on validation so there may be some clean up left to do


chipsec/hal/uefi_common.py
```diff
@@ -1199,8 +1199,8 @@ def __str__(self):
         if self.unknown is not None: str_r += "  Unknown: 0x{:04X}\n".format(self.unknown)
         if self.count   is not None: str_r += "  Count  : 0x{:X}\n".format(self.count)
         if self.values  is not None:
-            fmt = '0x%0' + ( '{:d}X'.format(script_width_sizes[self.width]*2) )
-            str_r += "  Values : %s\n" % ("  ".join( [fmt % v for v in self.values] ))
+            fmt = '0x{}'.format( '{}{:d}X{}'.format('{:',scrpt_width_sizes[self.width]*2,'}') )
```
frinzell: This can be simplified to:
```python
fmt = '0x{}'.format( '{{{:d}X}}'.format(scrpt_width_sizes[self.width]*2) )
```

chipsec/hal/uefi_common.py
```diff
@@ -1733,7 +1733,7 @@ class EFI_CONFIGURATION_TABLE():
     def __init__( self ):
         self.VendorTables = {}
     def __str__(self):
-        return ( 'Vendor Tables:\n%s' % (''.join( ['{%s} : 0x{:016X}\n' % (vt,self.VendorTables[vt]) for vt in self.VendorTables])) )
+        return ( 'Vendor Tables:\n{}'.format(''.join( ['{}{}{} : 0x{:016X}\n'.format('{',vt,'}',self.VendorTables[vt]) for vt in self.VendorTables])) )
```
frinzell: This can be simplified to:
```python
return ( 'Vendor Tables:\n{}'.format(''.join( ['{{{}}} : 0x{:016X}\n'.format(vt,self.VendorTables[vt]) for vt in self.VendorTables])) )
```

ErikBjorge:
This needs to have a rebase done on it or dropped for a new version of the change.  Smaller chunks would be nice as well.

chipsec/hal/uefi_common.py
BrentHoltsclaw: Done

chipsec/hal/uefi_common.py
BrentHoltsclaw: Done

ErikBjorge:
I would like to merge this to the development branch with the existing CI failures.  They will need to be resolved prior to merging into master.  @frinzell, can you please take a look at the latest version since you have requested changes?

PR #437, opened 2018-09-12
Add SMBUS support for Denverton SoC

kingsumos:
Tested in UEFI environment, e.g. dump DDR SPD data:
python chipsec_util.py smbus read 0xa0 0x0 0x100


kingsumos:
Rework required to commit changes using a single git commit...

PR #442, opened 2018-09-21
Python 3 Compatability changes

BrentHoltsclaw:


BrentHoltsclaw:
Not sure that I understand the concern.  String format changes are within a different pull request

ErikBjorge:
Will merge onto development branch so we can continue development of the Python 3 support.  Will need to fix the unit test code.

PR #451, opened 2018-10-06
Fix uefi decompress when NextFileFw gets a invalid filesize

BrentHoltsclaw:


chipsec/hal/uefi_common.py
```diff
@@ -640,16 +640,16 @@ def GetFvHeader(buffer, off = 0):
     if logger().DEBUG:
         logger().log({}.format(
```
PerezFederico: This does not seem right. 
It fails when run with -d

PerezFederico:
It looks good to me. I tested it with my test image and it seems to work fine.

Just the comment, it fails with debug options.

PerezFederico:
Seems good to me!

PR #454, opened 2018-10-11
Logging XML and JSON changes

BrentHoltsclaw:
Combined PR's 422, 426, 427.  Resolves all issues with merging the files.  Leaves output functionality the same.


BrentHoltsclaw:
Changes made.  Added a function to return results so that deltas switch works correctly.  Other was a spelling error.

PR #455, opened 2018-10-12
CFL SPI FREG base/limit uses 15 bits

GHswitt:
Coffeelake PCH uses 15 bits for SPI flash region base/limit registers


GHswitt:
Sorry, please use #459

PR #462, opened 2018-10-20
WIN10 driver update

BrentHoltsclaw:
Pushing as an alternative solution to other pull requests.  This one leaves old solution in place and adds new process to work with VS2017 and Windows 10 WDK.


ErikBjorge:
@marcosdiazr can you work with @BrentHoltsclaw to iron out some of the deltas between the two solutions?

marcosdiazr:
Ok,
@BrentHoltsclaw I have a couple of questions.
IIRC you can generate correct drivers for both win10 and win 7 (and all versions between), so I would say this is valid to all versions. (And set target version to windows 7)
 Why would you use KMDF mode? Does it work correctly since driver.c  initializes using WDM and not KMDF functions? I'd use a WDM mode solution.
my PR is with WDM and target version on windows 7 and it works ok on Windows 10.
Just to ease configurations Please remove ARM modes of compilling.
Also update Readme to say that this can be used

BrentHoltsclaw:
@marcosdiazr I went with KMDF because WDM does not appear to be supported anymore within the latest WDK.  My goal was to create a new solution to work with the lastest WDK and leave the old process and build instructions as is.

Agree with the ARM Modes logic.

marcosdiazr:
I thought that WDF(KMDF) was built on top of WDM. So I don't think they can deprecate it. Anyways chipsec doesn't use any KMDF function at all. Isn't it better to use a WDM project that will work for all versions of windows from 7 to 10?
I insist I was able to build driver on win 10 I don't recall if I have latest lates wdk but at least from this year.
My concerns are that the driver can be built the same for all versions of windows using VS solution and I think that can be possible with WDM, or maybe KMDF if you uses version 1.9 I think. Maybe change the Target version to windows 7 and set KMDF version to 1.9 that should work for all windows from 7

marcosdiazr:
Now I could test this branch, Release modes are broken, please also fix that:
for x64 I got\
```
1>driver.obj : error LNK2019: unresolved external symbol WdmlibIoCreateDeviceSecure referenced in function DriverEntry
1>driver.obj : error LNK2001: unresolved external symbol SDDL_DEVOBJ_SYS_ALL_ADM_ALL
```

you may need to add $(DDK_LIB_PATH)wdmsec.lib; to linker input for release

For x86 Release I got
```
1>i386\cpu.asm(3): fatal error A1000: cannot open file : callconv.inc
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\VC\VCTargets\BuildCustomizations\masm.targets(69,5): error MSB3721: The command "ml.exe /c /nologo /Zi /Fo"Release\cpu.obj" /W3 /errorReport:prompt  /Tai386\cpu.asm" exited with code 1.
```

marcosdiazr:
So I'd:
 set Driver target version to win 7,
Fix Release builds
Erase ARM configurations

That should be it, I will close the other PR to avoid confussion

BrentHoltsclaw:
@marcosdiazr i made some of the requested changes however I was unable to set Driver target version to Win7 (can override however VS2017 ignores the edit)
@marcosdiazr @abazhaniuk @ErikBjorge how would you like to proceed?

marcosdiazr:
Go to Project->Properties
Ensure that in configuration you have: All Configurations and All platforms selected
Driver settings->General-> TargetOS Version set to Windows 7
Target Platform set to Desktop

BrentHoltsclaw:
Thanks for the pointer

marcosdiazr:
Sorry to insist, but why creating a separate config for win7 if that works for 10 also? or am I wrong?
And also, x64 versions copy the driver to chipsec\helper\win\win7_x64 but chipsec looks for the driver at chipsec\helper\win\win7_**amd**64

BrentHoltsclaw:
Thanks for the feedback, can you please do another check

marcosdiazr:
Checked, and Debug x64 is still copying it to win7_x64 folder instead of win7_amd64 folder. Release-x64 seems fixed.
When you want to do changes for all configurations, take care of selecting all configurations in the upper tab.
Everything else seems to be working ok

marcosdiazr:
It's OK, @ErikBjorge can you merge?

PR #471, opened 2018-11-03
Add Record/Replay ability into chipsec_main

BrentHoltsclaw:
Specify --clone within chipsec_main and it will write out the helper accesses into a json file.  Afterwards the file can be used to replay the event
Added a generic file helper which will load register accesses from a json file and use those instead of a driver.

### DIFF
--- a/chipsec/helper/file/helper.py
+++ b/chipsec/helper/file/helper.py
@@ -37,14 +37,18 @@ class FileCmds:
             self.filename = filename
 
     def AddElement(self,cmd,args,ret):
+        try:
+            margs = '({})'.format(','.join(str(i) for i in args))
+        except:
+            margs = str(args)
         if self.data.has_key(str(cmd)):
-            if self.data[str(cmd)].has_key(str(args)):
+            if self.data[str(cmd)].has_key(margs):
                 #using insert opposed to append so that it creates last in first out when using pop command within getElement
-                self.data[str(cmd)][str(args)].insert(0,ret)
+                self.data[str(cmd)][margs].insert(0,ret)
             else:
-                self.data[str(cmd)][str(args)] = [ret] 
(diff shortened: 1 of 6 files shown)


BrentHoltsclaw:
Made significant changes from initial please review and provide feedback.

frinzell:
I think clarifying the command line arguments, use cases, and the EFI helper fix remain.

chipsec/helper/file/helper.py
```diff
@@ -0,0 +1,297 @@
...
+
+class FileCmds:
+    def __init__(self, filename):
+        self.data = {}
+        if filename == "":
+            self.filename = "clone.json"
+        else:
+            self.filename = filename
+
+    def AddElement(self,cmd,args,ret):
```
johnloucaides: Can we do something like push/pop cmd/args/ret so that when we read/write the same address multiple times, we save the various values in order? I don't think this should be blocking acceptance, but it would probably avoid problems later to have it.

skochinsky:
I'm late to the party but just wanted to comment that  "clone"  seems to be a really non-obvious switch to me. I'd rather expect something more in the vein of "record/replay" terms.

chipsec/helper/file/helper.py
BrentHoltsclaw: Added a list to AddElement, which should resolve the concern.  Note that insert() was used opposed to append() to get first in first out when using pop() within getElement. Let me know if other changes are necessary.

johnloucaides:
Suggested changes appear to be reflected in the latest commits. I really like the idea of this capability. Are we looking for anything more to get it merged?

ErikBjorge:
Just need some conflicts to be cleaned up before we can merge.

PR #472, opened 2018-11-05
Documentation updates

ErikBjorge:
Updates to build scripts and documentation contents.  This should be a more accurate reflection of the current project state.


BrentHoltsclaw:
Page 2: Step 5 is within the drivers\win7\readme.  Also with VS Builds it will be done for you now

Page 3: spacing between step 6 and sub bullet one

Page 5: Note for linux installation is no longer applicable

Page 5: OS X, alternate to disable sip - "csrutil enable --without kext" will keep SIP enabled on Intel macs while allowing untrusted/unsigned kexts to load

BrentHoltsclaw:
Do we want to comment on unit testing? (how to test without driver build)

Issue #479, opened 2018-11-22
chipec_util uefi decode crash


0xfede7c8:
With the example provided below I'm getting:

``` bash
➜  sudo chipsec_util uefi decode IMAGE9K.ROM
################################################################
##                                                            ##
##  CHIPSEC: Platform Hardware Security Assessment Framework  ##
##                                                            ##
################################################################
[CHIPSEC] Version 1.3.6
****** Chipsec Linux Kernel module is licensed under GPL 2.0
[CHIPSEC] API mode: using CHIPSEC kernel module API
[CHIPSEC] Executing command 'uefi' with args ['decode', 'IMAGE9K.ROM']

[CHIPSEC] Parsing EFI volumes from 'IMAGE9K.ROM'..
Traceback (most recent call last):
  File "/usr/local/bin/chipsec_util", line 11, in <module>
    load_entry_point('chipsec==1.3.6', 'console_scripts', 'chipsec_util')()
  File "/usr/local/lib/python2.7/dist-packages/chipsec-1.3.6-py2.7-linux-x86_64.egg/chipsec_util.py", line 261, in main
    return chipsecUtil.main()
  File "/usr/local/lib/python2.7/dist-packages/chipsec-1.3.6-py2.7-linux-x86_64.egg/chipsec_util.py", line 228, in main
    comm.run()
  File "/usr/local/lib/python2.7/dist-packages/chipsec-1.3.6-py2.7-linux-x86_64.egg/chipsec/utilcmd/uefi_cmd.py", line 243, in run
    decode_uefi_region(_uefi, cur_dir, filename, fwtype)
  File "/usr/local/lib/python2.7/dist-packages/chipsec-1.3.6-py2.7-linux-x86_64.egg/chipsec/hal/spi_uefi.py", line 533, in decode_uefi_region
    parse_uefi_region_from_file( _uefi, fname, fwtype, fv_pth )
  File "/usr/local/lib/python2.7/dist-packages/chipsec-1.3.6-py2.7-linux-x86_64.egg/chipsec/hal/spi_uefi.py", line 515, in parse_uefi_region_from_file
    tree = build_efi_model(_uefi, rom, fwtype)
  File "/usr/local/lib/python2.7/dist-packages/chipsec-1.3.6-py2.7-linux-x86_64.egg/chipsec/hal/spi_uefi.py", line 400, in build_efi_model
    model = build_efi_tree( _uefi, data, fwtype )
[...]

PR #483, opened 2018-11-27
Fix issue #479

BrentHoltsclaw:
Added error handling for when EFI_SECTION_GUID_DEFINED is not large enough to contain dataoffset and attributes


0xfede7c8:
What happens with the section's data when this happens? Is the data retrievable or valid?

BrentHoltsclaw:
The error was occurring because the section image is not large enough to extract the guid information, DataOffset, and attributes.  The section data is essentially non-existent.  I put in a warning for the logger to report the inconsistency as something is not correct.  sec.DataOffset is only used if the sec.Guid matches one of the compression guids.  I created a fake guid of 0000-00-00-0000 to ensure that no errors are created and decompression will continue.

0xfede7c8:
Thanks for the clarification! Another one: Then, that fake guid and broken section will show up somewhere? In the json or lst files maybe?

BrentHoltsclaw:
It will show up at the top of the .lst file as

> WARNING: EFI Section seems to be malformed

PR #484, opened 2018-11-30
Use argparse within chipsec_main

BrentHoltsclaw:


chipsec/testcase.py
```diff
@@ -36,6 +36,24 @@ class ExitCode:
     INFORMATION   = 64
     NOTAPPLICABLE = 128
 
+    Help_String = """\
```
ErikBjorge: Can we name this help_epilog to be a little more clear on how it is used.

chipsec_main.py
```diff
@@ -373,117 +356,49 @@ def run_all_modules(self):
...
-                self.IMPORT_PATHS.append(a)
-            elif o in ("--failfast"):
-                self.failfast = True
-            elif o in ("--no_time"):
-                self.no_time = True
-            elif o in ("--deltas"):
-                self._deltas_file = a
-            else:
-                assert False, "unknown option"
+        parser = argparse.ArgumentParser(prog='chipsec_main.py',usage='%(prog)s [options]',formatter_class=argparse.RawDescriptionHelpFormatter,epilog=ExitCode.Help_String,add_help=False)
```
ErikBjorge: Do we want to leave the .py off the program name.  This may be better for people that install it on a test image.

BrentHoltsclaw:
@ErikBjorge thanks for the suggestions.  I made the fixes and squashed them.  

I also changed the ordering so that we do not try to do anything with the chipset or helpers before parsing arguments.

PR #485, opened 2018-11-30
Use argparse within chipsec_util

BrentHoltsclaw:


chipsec_util.py
```diff
@@ -124,40 +98,32 @@ def f_mod(self, x):
...
-                self._unkownPlatform = False
-                logger().log( "[*] Ignoring unsupported platform warning and continue execution" )
-            elif o in ("-l", "--log"):
-                self.set_logfile(a)
-            elif o in ("-n", "--no_driver"):
-                self._no_driver = True
-            else:
-                pass
-
+        parser = argparse.ArgumentParser(prog='chipsec_util.py',usage='%(prog)s [options] <command>',add_help=False)
```
ErikBjorge: Do we want to drop the .py from the prog paramter?

BrentHoltsclaw: Agree and changed

PR #495, opened 2018-12-19
Remove metaclass from oshelpers.py

BrentHoltsclaw:


ErikBjorge:
Merging into development branch so we can continue Python 3 development.

Issue #186, opened 2017-03-20
chipsec support for legacy BIOS and UEFI-compatibility mode boot.


ismaws:
Currently, chipsec efi symbol is missing on OS with BIOS and UEFI-compatibility mode boot. We need to workaround the efi structure to compile the driver successfully.

chipsec driver build errors:

WARNING: "efi" [/home/user/git/chipsec/build/lib.linux-x86_64-2.7/drivers/linux/chipsec.ko] undefined!

abazhaniuk:
what version of chipsec did you test? 
what version OS and kernel you have? 
i've tested latest chipsec on kali with legacy BIOS and chipsec works fine.

DropBearWare:
I think I'm hitting the same deal.

-- build log --
`root@ubuntu:/home/sds/chipsec# python setup.py build_ext -i
running build_ext
copying build/lib.linux-x86_64-2.7/chipsec/helper/linux/cores.so -> chipsec/helper/linux
copying build/lib.linux-x86_64-2.7/chipsec_tools/efi_compressor.so -> chipsec_tools
building the linux driver
creating /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers
creating /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux
copying drivers/linux/COPYING -> /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux
copying drivers/linux/Makefile -> /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux
copying drivers/linux/dkms.conf -> /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux
copying drivers/linux/AUTHORS -> /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux
creating /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux/include
copying drivers/linux/include/chipsec.h -> /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux/include
copying drivers/linux/WARNING.txt -> /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux
copying drivers/linux/README -> /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux
creating /home/sds/chipsec/build/lib.linux-x86_64-2.7/drivers/linux/i386
[...]

PR #500, opened 2019-01-03
Linux driver support for legacy BIOS and UEFI-compatibility mode

BrentHoltsclaw:
Issue #186

### DIFF
--- a/drivers/linux/chipsec_km.c
+++ b/drivers/linux/chipsec_km.c
@@ -28,10 +28,13 @@ chipsec@intel.com
 #include <linux/slab.h>
 #include <asm/io.h>
 #include <linux/smp.h>
-#include <linux/efi.h>
 #include <linux/miscdevice.h>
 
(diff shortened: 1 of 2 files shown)


drivers/linux/chipsec_km.c
```diff
@@ -702,10 +705,9 @@ void * patch_read_msr(void * CPUInfo)
 	return NULL;
 }
 
-
+#ifdef EFI_NOT_READY
```
ErikBjorge: Not sure why the chose EFI_NOT_READY as the magic value to check to see if efi.h was included.  Makes for some odd code to read.

BrentHoltsclaw: That code was already within the driver I just moved it above the function given everything it does depends upon efi functionality.

ErikBjorge: Understood, just commenting on the original choice.  Nothing for you to change around this code.

Issue #249, opened 2017-06-26
Kernel module IOCTL numbers should be generated with <ioctl.h> macros


nickmooney:
Best practices for kernel modules dictate that IOCTL numbers should be generated with the _IO, _IOR, _IOW, _IOWR macros rather than laid down sequentially from 0.

This is partially for versioning purposes and partially to ensure that sending an IOCTL to the wrong device does not silently succeed.

nickmooney:
Looking at this issue further, it seems the IOCTL numbers *are* defined using the correct macros (in `drivers/linux/include/chipsec.h`), but without a unique parameter given as the first argument to the macro, giving the same result as if they were not used.

BrentHoltsclaw:
This should be fixed and merged within #506.  Please reopen the issue if there are farther concerns.  Thanks

PR #506, opened 2019-01-07
Address issue #249

BrentHoltsclaw:
changed drivers/linux/include/chipsec.h to start from 'C'
added logic to compute ioctl base based upon changes to chipsec.h above


drivers/linux/include/chipsec.h
```diff
@@ -24,28 +24,28 @@ chipsec@intel.com
...
-#define IOCTL_GET_EFIVAR               _IOWR(0, 0xD, int*)
-#define IOCTL_SET_EFIVAR               _IOWR(0, 0xE, int*)       
-#define IOCTL_RDCR                     _IOWR(0, 0x10, int*) 
-#define IOCTL_WRCR                     _IOWR(0, 0x11, int*) 
-#define IOCTL_RDMMIO                   _IOWR(0, 0x12, int*)
-#define IOCTL_WRMMIO                   _IOWR(0, 0x13, int*)
-#define IOCTL_VA2PA                    _IOWR(0, 0x14, int*)
-#define IOCTL_MSGBUS_SEND_MESSAGE      _IOWR(0, 0x15, int*)
-#define IOCTL_FREE_PHYSMEM             _IOWR(0, 0x16, int*)
+#define IOCTL_BASE                     _IO('C', 0)
```
ErikBjorge: Do we want a #define for the 'C' just in case we need/want to change it in the future?

BrentHoltsclaw: No problem

Issue #386, opened 2018-05-15
linux helper native (no driver) mmio read returns incorrect values when memory is under 4G


BrentHoltsclaw:
def native_read_mmio_reg(self, phys_address, size):
        if self.devmem_available():
            region = self.memory_mapping(phys_address, size)
            if not region:
                os.lseek(self.dev_mem, phys_address, os.SEEK_SET)
                reg = os.read(self.dev_mem, size)
            else:
                region.seek(phys_address - region.start)
                reg = region.read(size)
            return defines.unpack1(reg, size)

BrentHoltsclaw:
Closing as PR #507 should fix the issue.

PR #507, opened 2019-01-09
Address issue #386

BrentHoltsclaw:
Found that native linux helper needs to first map io memory before reading it
    -  changed hal and helper functionality to allow base address and offset to be passed
  to the function allowing the call to native_map_io_space to be used without guessing the base address
Discovered a bug within hal list_MMIO_BARs that failed if a MMIOBAR was unreadable 
    -  added a try except to fix

### DIFF
--- a/chipsec/hal/mmio.py
+++ b/chipsec/hal/mmio.py
@@ -222,7 +222,6 @@ class MMIO(hal_base.HALBase):
     #
     def read_MMIOBAR_reg(self, bar_id, offset ):
         bar_base  = self.MMIO_BAR_base[ bar_id ]
-        #reg_addr  = bar_base + offset
         reg_value = self.cs.helper.read_mmio_reg( bar_base, 4, offset )
         if logger().VERBOSE:
             logger().log( '[mmio] %s + 0x%08X (0x%08X) = 0x%08X' % (MMIO_BAR_name[bar_id], offset, reg_addr, reg_value) )
--- a/chipsec/hal/mmio.py
+++ b/chipsec/hal/mmio.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2018, Intel Corporation
+#Copyright (c) 2010-2019, Intel Corporation
 #
 #This program is free software; you can redistribute it and/or
 #modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


chipsec/hal/mmio.py
```diff
@@ -222,8 +222,8 @@ def get_MMIO_base_address(self, bar_id):
     #
     def read_MMIOBAR_reg(self, bar_id, offset ):
         bar_base  = self.MMIO_BAR_base[ bar_id ]
-        reg_addr  = bar_base + offset
-        reg_value = self.cs.helper.read_mmio_reg( reg_addr, 4 )
+        #reg_addr  = bar_base + offset
```
ErikBjorge: Remove dead code

chipsec/utilcmd/mmcfg_cmd.py
```diff
@@ -53,7 +53,7 @@ def run(self):
 
         if 2 == len(self.argv):
             #pciexbar = _mmio.get_PCIEXBAR_base_address()
-            pciexbar = _mmio.get_MMCFG_base_address()
+            pciexbar, pciexbar_sz = _mmio.get_MMCFG_base_address()
             self.logger.log( "[CHIPSEC] Memory Mapped Config Base: 0x%016X" % pciexbar )
```
ErikBjorge: Add the size of the BAR here for debug purposes.

chipsec/helper/linux/helper.py
```diff
@@ -547,33 +547,34 @@ def read_mmio_reg(self, phys_address, size):
...
-            region = self.memory_mapping(phys_address, size)
+            region = self.memory_mapping(bar_base, bar_size)
             if not region:
-                os.lseek(self.dev_mem, phys_address, os.SEEK_SET)
-                reg = os.read(self.dev_mem, size)
-            else:
-                region.seek(phys_address - region.start)
-                reg = region.read(size)
+                self.native_map_io_space(bar_base, bar_size)
+                region = self.memory_mapping(bar_base, bar_size)
```
ErikBjorge: Couple of questions:
Have you tested cases where bar_size is None and you access the region (updating mapping) multiple times?  An example would be reading offset 0 and then offset 0x10.
Are we always able to map a region or do we need some error handling in case after mapping the IO space in case memory_mapping returns None?

BrentHoltsclaw: I have tested where bar_size is None and did not encounter any problems.  The dump command does this by default.  
Added an error message if the region does not map.

PR #512, opened 2019-01-14
Removing python brew command for MacOS

ErikBjorge:
It seems that python 2 is already installed on MacOS builds.


ErikBjorge:
Normally would let this sit for further review but CI should work.

Issue #306, opened 2017-10-19
Chipsec error on Legacy Win7 OS


BigBigNemo:
May i knon if this tool support on legacy Win7 OS? We encounter some errors when run this tool on Legacy Win7 OS, no problem with UEFI W7 or W10 OS. Thanks.

ERROR: chipsec.modules.common.secureboot.variables
ERROR: chipsec.modules.common.uefi.access_platform
ERROR: chipsec.modules.common.uefi.access_uefispec

skochinsky:
The cited modules require UEFI mode, so it's normal that they fail in legacy boot.

ErikBjorge:
These modules should be skipped if run on a system booting in legacy mode.

BrentHoltsclaw:
Addressed within PR #515.  Please reopen if there are farther concerns.

PR #514, opened 2019-01-16
detect legacy BIOS on win7+    fix for issue #306

NicholasArmour:
https://gallery.technet.microsoft.com/scriptcenter/Determine-UEFI-or-Legacy-7dc79488

For Windows 7/Server 2008R2 and above, the GetFirmwareEnvironmentVariable Win32 API (designed to extract firmware environment variables) can be used.  This API is not supported on non-UEFI firmware and will fail in a predictable way when called - this will identify a legacy BIOS.  On UEFI firmware, the API can be called with dummy parameters, and while it will still fail (probably!) the resulting error code will be different from the legacy BIOS case.

### DIFF
--- a/chipsec/helper/win/win32helper.py
+++ b/chipsec/helper/win/win32helper.py
@@ -723,8 +723,8 @@ class Win32Helper(Helper):
     # EFI Variable API
     #
     def EFI_supported( self):
-        # @TODO: use GetFirmwareType ?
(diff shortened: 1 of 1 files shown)


BrentHoltsclaw:
From the link call GetFirmwareEnvironmentVariableA() 
then check for an error (suggest using` win32api.GetLastError()`.  If 1 return False else return True.

PR #515, opened 2019-01-17
detect legacy BIOS on win7+ fix for issue #306

NicholasArmour:
https://gallery.technet.microsoft.com/scriptcenter/Determine-UEFI-or-Legacy-7dc79488

For Windows 7/Server 2008R2 and above, the GetFirmwareEnvironmentVariable Win32 API (designed to extract firmware environment variables) can be used. This API is not supported on non-UEFI firmware and will fail in a predictable way when called - this will identify a legacy BIOS. On UEFI firmware, the API can be called with dummy parameters, and while it will still fail (probably!) the resulting error code will be different from the legacy BIOS case.

### DIFF
--- a/chipsec/helper/win/win32helper.py
+++ b/chipsec/helper/win/win32helper.py
@@ -723,9 +723,15 @@ class Win32Helper(Helper):
     # EFI Variable API
     #
     def EFI_supported( self):
-        # kern32.GetFirmwareEnvironmentVariable with garbage parameters will return error = 1 reliably on a legacy system
-        self.GetFirmwareEnvironmentVariable("","{00000000-0000-0000-0000-000000000000}",0,0)
-        return win32api.GetLastError() !=1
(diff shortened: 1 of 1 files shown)


chipsec/helper/win/win32helper.py
```diff
@@ -723,8 +723,9 @@ def get_descriptor_table( self, cpu_thread_id, desc_table_code  ):
     # EFI Variable API
     #
     def EFI_supported( self):
-        # @TODO: use GetFirmwareType ?
-        return ((self.GetFirmwareEnvironmentVariable is not None) or (self.GetFirmwareEnvironmentVariableEx is not None))
+        # kern32.GetFirmwareEnvironmentVariable with garbage parameters will return error = 1 reliably on a legacy system
+        self.GetFirmwareEnvironmentVariable("","{00000000-0000-0000-0000-000000000000}",0,0)
```
ErikBjorge: Need to make sure the interface is not None prior to calling.  Looks like it may be None in some cases.

PR #518, opened 2019-01-29
Fixed 64-bit PA issue in UEFI Shell

ErikBjorge:
Updated Python UEFI Shell support to enable passing 64-bit addresses when accessing memory/MMIO.


ErikBjorge:
I will rebuild the UEFI Python interpreters once we have all the UEFI Python changes done.  I can also hold off on merging the changes until they are all in place and ready.

ErikBjorge:
I have found a couple more issues so I will need to do a better fix.  Would like to hold off on merging this version.

BrentHoltsclaw:
There has been too much time between this fix and proposed new changes.  Merging this version and will address any father changes as necessary.

PR #524, opened 2019-02-02
Compression changes

BrentHoltsclaw:
1. Removed previous compression methodology
      a. All files were removed
    2. Using compression tools directly from tianocore/edk2/BaseTools
      a. Downloaded from master branch
      b. Changed Makefiles to build within Chipsec
      c. Added BrotliCompress
    3. Reworked Windows and Linux Helper to utilize tools built from compression tools
      a. Both helpers utilize same logic of calling compression tools
      b. Updated to support BrotliCompress
    4. Modified GUID functionality within uefi_common to unpack struct as characters
    5. TODO: need farther validation possibly some fixup with the build scripts


johnloucaides:
any idea what's up with the travis ci fail?

ErikBjorge:
Not sure at this point but I think it is a problem with the build scripts.  @BrentHoltsclaw will continue to work on this.  We will need to hold off on merging until the CI passes.

BrentHoltsclaw:
CI was broken due to an issue with the Linux helper, it should be fixed now

chipsec/helper/oshelper.py
```diff
@@ -100,13 +99,13 @@ class OsHelper:
     def __init__(self):
         self.helper = None
         self.loadHelpers()
-        #print "Operating System: %s %s %s %s" % (self.os_system, self.os_release, self.os_version, self.os_machine)
+        #print "Operating System: {} {} {} {}".format(self.os_system, self.os_release, self.os_version, self.os_machine)
```
ErikBjorge: Why are we fixing dead code.  Please remove the dead code.  It happens in a number of locations in this file.

chipsec/hal/uefi_platform.py
```diff
@@ -521,17 +521,17 @@ def __str__(self):
         return """
 Header (VSS)
 ------------
-VendorGuid : {%08X-%04X-%04X-%04s-%06s}
+VendorGuid : {%s}
```
ErikBjorge: As part of this change is you have fixed some print formatting and left other formatting.  Is this on purpose?

BrentHoltsclaw: The fixes were to help simplify some of the guid parsing and make it python 2/3 compliant.  Probably should have put them in a separate pull request...

chipsec/helper/oshelper.py
BrentHoltsclaw: ok will remove

Issue #523, opened 2019-02-01
Chipsec SMM modules working only on config options in every other arguments


bikashdash:
Hey guys,
 I am working on chipsec from quite sometime. recently i updated the latest release . 
It seems that, for every command line arguments, it i s only going on config options. if you see below i have choosen fuzz option in the command line. but evevn though it is hitting on config options

Anyone having same issue here?

C:\Users\Cr4sh\Documents\CPG_Work\chipsec>python chipsec_main.py -m tools.smm.smm_ptr -a fuzz,0:0x40
################################################################
################################################################
[CHIPSEC] Version 1.3.6
[CHIPSEC] Arguments: -m tools.smm.smm_ptr -a fuzz,0:0x40

WARNING: *******************************************************************
WARNING: Chipsec should only be used on test systems!
WARNING: It should not be installed/deployed on production end-user systems.
WARNING: See WARNING.txt
WARNING: *******************************************************************

[CHIPSEC] API mode: using CHIPSEC kernel module API
[CHIPSEC] OS      : Windows 10 10.0.17763 AMD64
[CHIPSEC] Platform: Mobile 7th Generation Core Processor (Kabylake H)
[CHIPSEC]      VID: 8086
[CHIPSEC]      DID: 5910
[CHIPSEC] PCH     : Default PCH
[CHIPSEC]      VID: 8086
[CHIPSEC]      DID: A154

[+] loaded chipsec.modules.tools.smm.smm_ptr
[*] running loaded modules ..

[*] running module: chipsec.modules.tools.smm.smm_ptr
[*] Module arguments (1):
['fuzz,0:0x40']
[x][ =======================================================================
[x][ Module: A tool to test SMI handlers for pointer validation vulnerabilies
[x][ =======================================================================
Usage: chipsec_main -m tools.smm.smm_ptr [ -a <mode>,<config_file>|<smic_start:smic_end>,<size>,<address> ]
  mode          SMI handlers testing mode
    = config    use SMI configuration file <config_file>
    = fuzz      fuzz all SMI handlers with code in the range <smic_start:smic_end>
    = fuzzmore  fuzz mode + pass '2nd-order' pointers within buffer to SMI handlers
  size          size of the memory buffer (in Hex)
  address       physical address of memory buffer to pass in GP regs to SMI handlers (in Hex)
    = smram     pass address of SMRAM base (system may hang in this mode!)

hello fuzz,0:0x40
[...]

PR #525, opened 2019-02-04
Fix issue #523.

BrentHoltsclaw:
When argparse was added to chipsec_main it no longer requires -a arguements to be separated with a comma.
Adding fix into chipsec main to handle cases when comma separated values are given
@TODO with different request:
Update any test docs to remove the comma within arguements for the modules

### DIFF
--- a/chipsec_main.py
+++ b/chipsec_main.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2018, Intel Corporation
+#Copyright (c) 2010-2019, Intel Corporation
 #
 #This program is free software; you can redistribute it and/or
 #modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


chipsec_main.py
```diff
@@ -399,6 +399,9 @@ def parse_args(self):
             logger().DEBUG   = True
         if self.log:
             logger().set_log_file( self.log )
+        if self._module_argv and len(self._module_argv) == 1 and len(self._module_argv[0].split(',')) > 1:
```
ErikBjorge: You should be using `self._module_argv[0].count(',') >= 1` instead of splitting and generating a new list.

chipsec_main.py
```diff
@@ -399,7 +399,7 @@ def parse_args(self):
             logger().DEBUG   = True
         if self.log:
             logger().set_log_file( self.log )
-        if self._module_argv and len(self._module_argv) == 1 and len(self._module_argv[0].split(',')) > 1:
+        if self._module_argv and len(self._module_argv) == 1 and len(self._module_argv[0].count(',')) >= 1:
```
skochinsky: `count()` already returns the number of commas, no need to call `len()` on it. And since you only need to check for _one_ comma, something like `',' in self._module_argv[0]` is even better IMO.

ErikBjorge: Agree, I missed the len call still being in the commit.

chipsec_main.py
```diff
@@ -399,6 +399,9 @@ def parse_args(self):
             logger().DEBUG   = True
         if self.log:
             logger().set_log_file( self.log )
+        if self._module_argv and len(self._module_argv) == 1 and self._module_argv[0].count(','):
+            logger().log( "[*] Use of the -a command no longer needs to have arguments concatinated with ','")
```
mjauregu: Small typo: concatinated -> concatenated. Otherwise, looks good

PR #541, opened 2019-03-09
Update chipsec_util acpi_cmd to argparse

BrentHoltsclaw:
Requires PR #540

### DIFF
--- a/chipsec/utilcmd/acpi_cmd.py
+++ b/chipsec/utilcmd/acpi_cmd.py
@@ -60,6 +60,8 @@ class ACPICommand(BaseCommand):
         parser_table.add_argument('_name',metavar='table|filename',nargs=1,help="table to list")
         parser_table.set_defaults(func=self.acpi_table)
         parser.parse_args(self.argv[2:],namespace=ACPICommand)
+        if self.func == self.acpi_table and self._file:
+            return False
         return True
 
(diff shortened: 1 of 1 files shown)


chipsec/utilcmd/acpi_cmd.py
```diff
@@ -50,49 +51,43 @@ class ACPICommand(BaseCommand):
     """
 
     def requires_driver(self):
-        # No driver required when printing the util documentation
-        if len(self.argv) < 3:
-            return False
```
ErikBjorge: This looks like we will always need the driver even to get help information.  Do we want to make a decision about loading the driver after we call parser.parse_args?

BrentHoltsclaw: Please re-review.  I put the argument parser within requires_driver function, therefore if there is a problem parsing the args or help is called within the command, it will show the help screen before returning.

ErikBjorge: Understood that it was called first but forgot that it would call sys.exit() and not return from requires_driver.  My bad.

So in the case of accepting a file as input we should not need to load the driver.  Can we handle this case in the pull request?

BrentHoltsclaw: Changes added, let me know what you think.

PR #542, opened 2019-03-21
Moving Helper class from oshelper.py to basehelper.py

BrentHoltsclaw:


chipsec/helper/basehelper.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def stop( self, start_driver ):
+        if logger().VERBOSE:
+            logger().log("[helper] File Helper stopped/unloaded")
+        return True
+
+    def delete(self, start_driver):
+        if logger().VERBOSE:
+            logger().log("[helper] File Helper deleted")
+        return True
```
ErikBjorge: Create, start, stop and delete have File Helper message and return True.  I think these should also raise exceptions so that derived classes MUST implement these functions.

BrentHoltsclaw: Changes are complete.  I left in the logging, do you want that gone too?

ErikBjorge: Logging is fine but it should not reference the File Helper.  You could have it say that the method was not implemented.

BrentHoltsclaw: Thanks!

PR #556, opened 2019-05-09
Add vulnerability check to verify that SPD Write Disable bit in SMBus controller is set

wrffrz:
According to Intel PCH datasheets, the SMBus controller has a "SPD Write Disable" bit as part of the 
Host Configuration (HCFG) register which prevents writes to well-known SPD SMBus addresses and
should be set by BIOS.

### DIFF
--- a/chipsec/modules/common/spd_wd.py
+++ b/chipsec/modules/common/spd_wd.py
@@ -1,0 +1,67 @@
+# CHIPSEC: Platform Security Assessment Framework
+# Copyright (c) 2019, Eclypsium, Inc.
+# Copyright (c) 2019, Intel Corporation
+# 
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; Version 2.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+This module checks that SPD Write Disable bit in SMBus controller has been set
+
+References:
+
+Intel 8 Series/C220 Series Chipset Family Platform Controller Hub datasheet
+Intel 300 Series Chipset Families Platform Controller Hub datasheet
+
+This module checks the following:
+
+    SMBUS_HCFG.SPD_WD
+
+The module returns the following results:
+
+    FAILED : SMBUS_HCFG.SPD_WD is not set
+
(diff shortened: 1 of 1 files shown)


chipsec/modules/common/spd_wd.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def check_spd_wd(self):
+        self.logger.start_test( "SPD Write Disable" )
+
+        spd_wd_res = ModuleResult.FAILED
+        spd_wd_reg = self.cs.read_register( 'SMBUS_HCFG' )
+        spd_wd = self.cs.get_register_field( 'SMBUS_HCFG', spd_wd_reg, 'SPD_WD' )
+
+        if 0 == spd_wd:
+            self.logger.log_failed_check( "SPD Write Disable is not set" )
```
ErikBjorge: You may want to check to see if any SPDs can be found behind the controller.  In some configurations (memory soldered down for example) SPDs may not be present and the lock would not be required.  The SMBus could be used for other purposes in those cases.

chipsec/modules/common/spd_wd.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self):
+        BaseModule.__init__(self)
+
+    def is_supported(self):
+        return self.cs.is_device_enabled( 'SMBUS' )
+
+    def check_spd_wd(self):
+        self.logger.start_test( "SPD Write Disable" )
+
+        spd_wd_res = ModuleResult.FAILED
```
ErikBjorge: BaseModule.res should be used to store the final result for the module.  Either use it here or assign self.res with the module result in the Run method.

chipsec/modules/common/spd_wd.py
wrffrz: Makes sense.  I'll update the PR to address this and the BaseModule.res thing.  If no SPDs are present, should we skip the test altogether or warn if SPD_WD is unset rather than fail the test?

ErikBjorge: If you don't find any SPDs I would just report the state of the bit.  This might be a good time to use the Informational message type.  Have no value judgement.  Let me know what you think.

wrffrz: I think my latest commit addresses this and the self.res issue.  Thanks for the feedback.

chipsec/modules/common/spd_wd.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def check_spd_wd(self):
+        self.logger.start_test( "SPD Write Disable" )
+
+        try:
+            _smbus = smbus.SMBus( self.cs )
+            _spd   = spd.SPD( _smbus )
+        except BaseException, msg:
+            print msg
+            return
```
frinzell: Recommend:
- Using the CHIPSEC logging functions for the `msg`
- Setting `self.res` to `ModuleResult.ERROR` before returning

wrffrz: Thanks for the feedback.  I think I've addressed these in the latest version of the patch.

ErikBjorge: Thanks, looks good.

@frinzell can you confirm and approve if all is good on your side?

PR #559, opened 2019-05-14
Add the eXecute permission on the launchable binaries

hughsie:
This means on Linux you can just do ./chipsec_util.py rather than using the
Python binary to invoke it.


hughsie:
@ErikBjorge what's the usual way to push these? Is there a bot merging PRs or do I need to do anything else?

ErikBjorge:
Merges are manual at this point.  I will take care of this one and a few others today and tomorrow.  Small changes like this normally land quickly.

Issue #561, opened 2019-05-14
UEFI decode explodes with TypeError: object of type 'bool' has no len()


hughsie:
To reproduce:

    wget https://fwupd.org/downloads/4689401be87da70d7ea5685c0941e42a4df40830-Lenovo-ThinkPad-P50-SystemFirmware-1.56.cab
    gcab -x 4689401be87da70d7ea5685c0941e42a4df40830-Lenovo-ThinkPad-P50-SystemFirmware-1.56.cab
    python2 ./chipsec_util.py -n uefi decode firmware.bin

    ################################################################
    ##                                                            ##
    ##  CHIPSEC: Platform Hardware Security Assessment Framework  ##
    ##                                                            ##
    ################################################################
    [CHIPSEC] Version 1.3.7
    [CHIPSEC] API mode: using OS native API (not using CHIPSEC kernel module)
    [CHIPSEC] Executing command 'uefi' with args ['decode', 'firmware.bin']

    [CHIPSEC] Parsing EFI volumes from 'firmware.bin'..
    Traceback (most recent call last):
      File "./chipsec_util.py", line 232, in <module>
        sys.exit( main() )
      File "./chipsec_util.py", line 228, in main
        return chipsecUtil.main()
      File "./chipsec_util.py", line 195, in main
        comm.run()
      File "~/Code/chipsec/chipsec/utilcmd/uefi_cmd.py", line 243, in run
        decode_uefi_region(_uefi, cur_dir, filename, fwtype)
      File "~/Code/chipsec/chipsec/hal/spi_uefi.py", line 551, in decode_uefi_region
        parse_uefi_region_from_file( _uefi, fname, fwtype, fv_pth )
      File "~/Code/chipsec/chipsec/hal/spi_uefi.py", line 533, in parse_uefi_region_from_file
        tree = build_efi_model(_uefi, rom, fwtype)
      File "~/Code/chipsec/chipsec/hal/spi_uefi.py", line 418, in build_efi_model
        model = build_efi_tree( _uefi, data, fwtype )
      File "~/Code/chipsec/chipsec/hal/spi_uefi.py", line 362, in build_efi_tree
        fwbin = build_efi_file_tree ( _uefi, fv_img, fwtype)
      File "~/Code/chipsec/chipsec/hal/spi_uefi.py", line 333, in build_efi_file_tree
[...]

PR #565, opened 2019-05-16
Decompress will return data not boolean when called through HAL

BrentHoltsclaw:
fix for #561


hughsie:
I confirm this fixed the bug -- thanks!

frinzell:
Just copyright

PR #566, opened 2019-05-16
Add dynamic bus support

frinzell:
- Allow for discovery of multiple/movable devices
- Allow for reading/writing/printing of multiple/movable devices
- Added XML attributes to chipsec_cfg.xsd

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -676,9 +676,9 @@ class Chipset:
             _did  = device.get( 'did', None )
             if (_vid and _did):
                 _bus = []
-                did  = [_.strip() for _ in _did.split(',')]
+                did  = [int(_,16) for _ in _did.split(',')]
                 for _dev in _devices:
-                    if ((device['dev'], device['fun'], _vid) == _dev[1:3]) and (_dev[4] in did):
+                    if ((int(device['dev'],16), int(device['fun'],16), int(_vid,16)) == _dev[1:4]) and (_dev[4] in did):
                         _bus.append( hex(_dev[0]) )
                         if logger().DEBUG: logger().log( '    + {:-16s}: VID 0x{:04X} - DID 0x{:04X} -> Bus 0x{:02X}'.format(_device, _dev[3], _dev[4], _dev[0]) )
                 if len(_bus):
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -561,6 +561,11 @@ class Chipset:
         if logger().DEBUG: logger().log("[*] Loading Configuration Files:")
         for _xml in loaded_files:
             self.init_cfg_xml(_xml, self.code, self.pch_code)
+
+        # Load Bus numbers for this platform.
+        if logger().DEBUG: logger().log("[*] Discovering Bus Configuration:")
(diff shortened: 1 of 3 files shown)


elicn:
My understanding in the CHIPSEC architecture is limited, but I am not sure this actually addresses the problem described. Servers may allocate different numbers to the same bus, and it may not be consistent across packages [e.g. B(1) is not necessarily 1, and might be assigned to 17 instead; another package on the same platform might assign it differently]. Accessing a PCI device should be done in respect of the thread / package in use, since the underlying bus number may be different.

frinzell:
It should address the immediate need.  In general, if a device is accessible, it should not matter what thread/package is used.   Can you elaborate on the specific corner case / concern?  Thanks!

chipsec/chipset.py
```diff
@@ -739,19 +763,28 @@ def is_register_defined(self, reg_name):
...
-        if reg_def["type"] == "pcicfg" and "device" in reg_def:
+        if (reg_def["type"] == "pcicfg" or reg_def["type"] == "mmcfg") and ("device" in reg_def):
             dev_name = reg_def["device"]
             if dev_name in self.Cfg.CONFIG_PCI:
                 dev = self.Cfg.CONFIG_PCI[dev_name]
                 reg_def['bus'] = dev['bus']
                 reg_def['dev'] = dev['dev']
                 reg_def['fun'] = dev['fun']
+                if dev_name in self.Cfg.BUS:
+                    reg_def['bus'] = self.Cfg.BUS[dev_name][bus_index]
```
BrentHoltsclaw: Is there a way to ensure that bus_index is a valid index?

chipsec/chipset.py
```diff
@@ -656,6 +661,23 @@ def init_cfg_xml(self, fxml, code, pch_code):
...
+        if logger().DEBUG: logger().log( '[*] loading device buses..' )
+        _devices = self.pci.enumerate_devices()
+        for _device in self.Cfg.CONFIG_PCI:
+            device = self.Cfg.CONFIG_PCI[_device]
+            _vid  = device.get( 'vid', None )
+            _did  = device.get( 'did', None )
+            if (_vid and _did):
+                _bus = []
+                did  = [int(_,16) for _ in _did.split(',')]
+                for _dev in _devices:
```
BrentHoltsclaw: the naming between _devices, _device, device, _dev is confusing, is there a better naming convention which would help?

chipsec/chipset.py
frinzell: Calling get_register_bus() or get_device_bus() will return the valid list.

BrentHoltsclaw: what if bus index is not in the range of `self.Cfg.BUS[dev_name][bus_index]`.  Please add a check to avoid an exception.

chipsec/chipset.py
frinzell: Due to overloading of the XML attributes and BDF.  I can add clarifying labels.

BrentHoltsclaw: thanks!

BrentHoltsclaw:
can you please resolve the conflicts with master

PR #570, opened 2019-05-17
Fix case issue within chipsec_tools LZMA directory name

BrentHoltsclaw:


BrentHoltsclaw:
@ErikBjorge I'm unable to reproduce the ci failure on a real system, how do you want to proceed?

PR #574, opened 2019-05-20
Add signed guid Certificate RSA2048/SHA256 support to UEFI decompress

BrentHoltsclaw:


frinzell:
Looks good.  Just needs copyright updates.

hughsie:
@BrentHoltsclaw can I ask what exactly these signatures represent? I'm unable to test this as the branch needs rebasing with recent fixes in master for the LVFS. I know there are a few people interested in "who signed these FV's" both in government and industrial fields.

ErikBjorge:
These specific GUIDs that indicate that a section is signed using a specific process.  The sections should be verified by the section extraction support in the firmware prior to use.  Just normal signature verification support.  Similar support exists for firmware volumes.

The reason we are adding it to CHIPSEC is that CHIPSEC would not extract the contents of these sections.  This caused us to not extract drivers from these sections and provide an incomplete inventory of the different firmware volume sections.

Currently we are just correctly identifying the section and extracting the contents.  In the future we could also verify the contents of these sections as well.  If PKCS7 is used we can even extract additional information.  I think all of this support will move us closer to being able to see who signed the contents and with what method.  It will also help us determine what content is not signed.  It will not help us determine if the content is correctly verified on boot or flash update.

All of this information can be found in the PI Specification or in the section extraction support in the SecurityPkg of EDKII.  I prefer code first and then documentation but it is up to you.

Source Example:
SecurityPkg\Library\DxeRsa2048Sha256GuidedSectionExtractLib\DxeRsa2048Sha256GuidedSectionExtractLib.c

PI Specification can be found here https://uefi.org/specifications (1.7 is the latest)
Volume 3 has the data you will want to look at.
* 3.2.1.1 - EFI Signed Firmware Volumes
* 3.2.5 - EFI Signed Sections

I hope this helps a little.  If I missed the point please let me know and I will try again to answer your question.

hughsie:
I get this when using `./chipsec_util.py uefi decode firmware.bin -n`:

      File "/home/hughsie/Code/chipsec/chipsec/hal/spi_uefi.py", line 245, in __str__
        if self.Comments: _s += "Comments {}".format(self.Comments)
    AttributeError: 'EFI_SECTION' object has no attribute 'Comments'

PR #578, opened 2019-05-27
C60X,C61X,C620 PCH updates

frinzell:
Requires PR #566 
- Update C60X PCH configuration to support dynamic BUS support
- Update C61X PCH configuration to support dynamic BUS support
- Update C620 PCH configuration to support dynamic BUS support

### DIFF
--- a/chipsec/cfg/pch_c620.xml
+++ b/chipsec/cfg/pch_c620.xml
@@ -39,7 +39,7 @@ XML configuration file for
   <!--                                      -->
   <!-- #################################### -->
   <io>
-    <bar name="ABASE"      register="ABASE"    base_field="BA"    size="0x100" desc="ACPI Base Address"/>
+    <bar name="ABASE"      register="ABASE"    base_field="Base"  size="0x100" desc="ACPI Base Address"/>
     <bar name="PMBASE"     register="ABASE"    base_field="BA"    size="0x100" desc="ACPI Base Address"/>
     <bar name="TCOBASE"    register="TCOBASE"  base_field="TCOBA" size="0x20"  desc="TCO Base Address"/>
 
@@ -122,6 +122,7 @@ XML configuration file for
       <field name="SMM_BWP"  bit="5" size="1" desc="SMM BIOS Write Protection" />
       <field name="BBS"      bit="6" size="1" desc="Boot BIOS Strap" />
       <field name="BILD"     bit="7" size="1" desc="BIOS Interface Lock Down"/>
+    </register>
 
     <!-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ -->
     <!-- MMIO registers               -->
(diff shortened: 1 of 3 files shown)


chipsec/cfg/pch_c620.xml
```diff
@@ -55,62 +59,69 @@ XML configuration file for
...
+    <register name="P2SB_HIDE"  type="pcicfg" device="P2SBC" offset="0xE1" size="1" desc="P2SB Configuration Register hide-unhide">
+      <field name="HIDE" bit="0" size="1" desc="Hide SBREG_BAR"/>
+    </register>
 
     <!-- Power Management Controller -->
-    <register name="ABASE" type="pcicfg" bus="0" dev="0x1f" fun="2" offset="0x40" size="4" desc="ACPI Base Address">
+    <register name="ABASE"       type="pcicfg" device="PMC" offset="0x40" size="4" desc="ACPI Base Address">
+      <field  name="Base" bit="8" size="24" desc="Base Address"/>
       <field name="STYPE" bit="0" size="1" desc="Space Type (always 1 - I/O space)"/>
-      <field name="BA"    bit="8" size="8" desc="Base Address"/>
```
iherna2: I think we need to keep the Base Address around, otherwise we are rendering the accesses to the BAR based registers useless

chipsec/cfg/pch_c620.xml
```diff
@@ -55,62 +59,69 @@ XML configuration file for
...
-    <register name="BC" type="pcicfg" bus="0" dev="0x1F" fun="5" offset="0xDC" size="4" desc="BIOS Control">
+    <register name="BC" type="pcicfg" device="SPI" offset="0xDC" size="4" desc="BIOS Control">
       <field name="BIOSWE"   bit="0" size="1" desc="BIOS Write Enable" />
       <field name="BLE"      bit="1" size="1" desc="BIOS Lock Enable" />
       <field name="SRC"      bit="2" size="2" desc="SPI Read Configuration" />
       <field name="TSS"      bit="4" size="1" desc="Top Swap Status" />
       <field name="SMM_BWP"  bit="5" size="1" desc="SMM BIOS Write Protection" />
       <field name="BBS"      bit="6" size="1" desc="Boot BIOS Strap" />
       <field name="BILD"     bit="7" size="1" desc="BIOS Interface Lock Down"/>
-    </register>
```
iherna2: This will make the xml file useless

chipsec/cfg/pch_c620.xml
frinzell: It's there, just named "Base" plus the size attribute is now fixed.   This aligns with what is in common.xml and what it overrides in the CPU xml for this platform.

chipsec/cfg/pch_c620.xml
frinzell: That should be there.

chipsec/cfg/pch_c620.xml
iherna2: I see that, then what we need to fix is the ABASE bar definition as it is pointing to field BA which is non-existent

Issue #589, opened 2019-06-04
chipsec_util decode fails,  Command '/usr/bin/chipsec_tools/compression/bin/LzmaCompress -d -o sect01_expg sect01_expg.gz' returned non-zero exit status 127


comicsheep:
It seems like there is no '/usr/bin/chipsec_tools/' directory with brand new installed chipsec 1.3.7

`[CHIPSEC] Decoding SPI ROM image from a file './H3C_fw_2019-06-03_172151.bin'
[CHIPSEC] Found SPI Flash descriptor at offset 0x0 in the binary './H3C_fw_2019-06-03_172151.bin'
/bin/sh: /usr/bin/chipsec_tools/compression/bin/LzmaCompress: No such file or directory
Traceback (most recent call last):
  File "/usr/bin/chipsec_util", line 9, in <module>
    load_entry_point('chipsec==1.3.7', 'console_scripts', 'chipsec_util')()
  File "/usr/lib64/python2.7/site-packages/chipsec-1.3.7-py2.7-linux-x86_64.egg/chipsec_util.py", line 228, in main
    return chipsecUtil.main()
  File "/usr/lib64/python2.7/site-packages/chipsec-1.3.7-py2.7-linux-x86_64.egg/chipsec_util.py", line 195, in main
    comm.run()
  File "/usr/lib64/python2.7/site-packages/chipsec-1.3.7-py2.7-linux-x86_64.egg/chipsec/utilcmd/decode_cmd.py", line 124, in run
    spi_uefi.decode_uefi_region(_uefi, pth, fname, fwtype)
  File "/usr/lib64/python2.7/site-packages/chipsec-1.3.7-py2.7-linux-x86_64.egg/chipsec/hal/spi_uefi.py", line 556, in decode_uefi_region
    parse_uefi_region_from_file( _uefi, fname, fwtype, fv_pth )
  File "/usr/lib64/python2.7/site-packages/chipsec-1.3.7-py2.7-linux-x86_64.egg/chipsec/hal/spi_uefi.py", line 538, in parse_uefi_region_from_file
    tree = build_efi_model(_uefi, rom, fwtype)
  File "/usr/lib64/python2.7/site-packages/chipsec-1.3.7-py2.7-linux-x86_64.egg/chipsec/hal/spi_uefi.py", line 423, in build_efi_model
    model = build_efi_tree( _uefi, data, fwtype )
[...]

PR #590, opened 2019-06-04
Fixup chipsec_tools path within linux helper

BrentHoltsclaw:
fix for issue #589


BrentHoltsclaw:
@hughsie, can you verify this doesn't break anything for you?  Thanks.

hughsie:
@BrentHoltsclaw it seems to work for me, the LVFS CI tests passd without issue.

Issue #557, opened 2019-05-10
chipsec_util spd detect doesn't print the detected SPDs without verbose flag


wrffrz:
Running "chipsec_util.py spd detect" doesn't actually print out the results of any SPDs that were detected unless you include the verbose flag, which results in way more detail than most people want.

> $ sudo python chipsec_util.py spd detect
> 
> ################################################################
> ##                                                            ##
> ##  CHIPSEC: Platform Hardware Security Assessment Framework  ##
> ##                                                            ##
> ################################################################
> [CHIPSEC] Version 1.3.7
> ****** Chipsec Linux Kernel module is licensed under GPL 2.0
> [CHIPSEC] API mode: using CHIPSEC kernel module API
> [CHIPSEC] Executing command 'spd' with args ['detect']
> 
> [CHIPSEC] Searching for DIMMs with SPD..
> [CHIPSEC] (spd) time elapsed 0.004
> $

It says that it's searching for DIMMs with SPD, but no results are printed.

However, if you add the -v flag, there's a lot more output including DIMMs that were detected:

> $ sudo python chipsec_util.py -v spd detect
> 
> ################################################################
> ##                                                            ##
> ##  CHIPSEC: Platform Hardware Security Assessment Framework  ##
> ##                                                            ##
> ################################################################
> [CHIPSEC] Version 1.3.7
> [CHIPSEC] Loaded command-line extensions:
>    ['io_cmd', 'vmem_cmd', 'ucode_cmd', 'chipset_cmd', 'spi_cmd', 'msgbus_cmd', 'spd_cmd', 'reg_cmd', 'mmcfg_cmd', 'mem_cmd', 'vmm_cmd', 'interrupts_cmd', 'mmio_cmd', 'deltas_cmd', 'tpm_cmd', 'msr_cmd', 'ec_cmd', 'igd_cmd', 'spidesc_cmd', 'smbus_cmd', 'cmos_cmd', 'acpi_cmd', 'cpu_cmd', 'iommu_cmd', 'uefi_cmd', 'pci_cmd', 'decode_cmd', 'desc_cmd']
> [helper] Linux Helper created
> Module /dev/chipsec loaded successfully
> ****** Chipsec Linux Kernel module is licensed under GPL 2.0
> [helper] Linux Helper started/loaded
> [CHIPSEC] API mode: using CHIPSEC kernel module API
> [pci] reading B/D/F: 0/0/0, offset: 0x00, value: 0x01008086
[...]

PR #593, opened 2019-06-05
fix for issue #557

NicholasArmour:


BrentHoltsclaw:
The logic within the hal was fine, recommend printing out the returned values within the util_cmd instead.

npmitche:
> The logic within the hal was fine, recommend printing out the returned values within the util_cmd instead.

I think I agree with Brent:
```
if( 'detect' == op ):
    self.logger.log( "[CHIPSEC] Searching for DIMMs with SPD.." )
    _dimms = _spd.detect()
    self.logger.log( "[CHIPSEC] Detected the following SPD devices:")
    for _dimm in _dimms: self.logger.log( "[CHIPSEC]     %s: 0x%02X".format(_spd.SPD_DIMMS[_dimm], _dimm) )
```

chipsec/hal/spd.py
```diff
@@ -364,7 +364,7 @@ def detect( self ):
         _dimms = []
         for d in SPD_DIMMS:
             if self.isSPDPresent( d ): _dimms.append( d )
-        if logger().HAL: 
+        if logger().HAL:
             logger().log( "Detected the following SPD devices:" )
             for _dimm in _dimms: logger().log( "%s: 0x%02X" % (SPD_DIMMS[_dimm],_dimm) )
```
npmitche: If you're modifying this file, might as well also update this from % to format?

ErikBjorge: Just a white space change so no need to do anything else.

PR #596, opened 2019-06-06
Fix SPD/SMBus related issues

ErikBjorge:
Fix exception when 'ecc' variable is not initialized due to DIMM detection issues.
Add the ability to do a software reset of the SMBus.

### DIFF
--- a/chipsec/hal/smbus.py
+++ b/chipsec/hal/smbus.py
@@ -105,6 +105,13 @@ class SMBus(hal_base.HALBase):
         cmd = self.cs.read_register( 'SMBUS_CMD' )
         if 0 == (cmd & 0x1): self.cs.write_register( 'SMBUS_CMD', (cmd|0x1) )
 
+    def reset_SMBus_controller(self):
+        reg_value = self.cs.read_register('SMBUS_HCFG')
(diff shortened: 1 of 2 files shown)


chipsec/hal/smbus.py
```diff
@@ -105,6 +105,13 @@ def enable_SMBus_host_controller( self ):
         cmd = self.cs.read_register( 'SMBUS_CMD' )
         if 0 == (cmd & 0x1): self.cs.write_register( 'SMBUS_CMD', (cmd|0x1) )
 
+    def reset_SMBus_controller(self):
+        reg_value = self.cs.read_register('SMBUS_HCFG')
+        self.cs.write_register('SMBUS_HCFG', reg_value | 0x08)
+        for i in range(SMBUS_POLL_COUNT):
+            if (self.cs.read_register('SMBUS_HCFG') & 0x08) == 0:
+                return True
```
frinzell: Probably won't see an issue reading multiple times like this but using a slight delay might be better.

ErikBjorge: The delay in performing the I/O should be enough in this case.

Issue #603, opened 2019-06-21
UEFI parsing issue


abazhaniuk:
```
> sudo python chipsec_util.py spi dump lenovo_thinkpad_x1_carbon_6.bin

################################################################
##                                                            ##
##  CHIPSEC: Platform Hardware Security Assessment Framework  ##
##                                                            ##
################################################################
[CHIPSEC] Version 1.3.7
****** Chipsec Linux Kernel module is licensed under GPL 2.0
[CHIPSEC] API mode: using CHIPSEC kernel module API
[CHIPSEC] Executing command 'spi' with args ['dump', 'lenovo_thinkpad_x1_carbon_6.bin']

[CHIPSEC] dumping entire SPI flash memory to 'lenovo_thinkpad_x1_carbon_6.bin'
[CHIPSEC] it may take a few minutes (use DEBUG or VERBOSE logger options to see progress)
[CHIPSEC] BIOS region: base = 0x00700000, limit = 0x00FFFFFF
[CHIPSEC] dumping 0x01000000 bytes (to the end of BIOS region)
[spi] reading 0x1000000 bytes from SPI at FLA = 0x0 (in 262144 0x40-byte chunks + 0x0-byte remainder)
[CHIPSEC] completed SPI flash dump to 'lenovo_thinkpad_x1_carbon_6.bin'
[CHIPSEC] (spi dump) time elapsed 53.079
> sudo python chipsec_util.py decode lenovo_thinkpad_x1_carbon_6.bin

################################################################
##                                                            ##
##  CHIPSEC: Platform Hardware Security Assessment Framework  ##
##                                                            ##
################################################################
[CHIPSEC] Version 1.3.7
[CHIPSEC] API mode: using OS native API (not using CHIPSEC kernel module)
[CHIPSEC] Executing command 'decode' with args ['lenovo_thinkpad_x1_carbon_6.bin']

[CHIPSEC] Decoding SPI ROM image from a file 'lenovo_thinkpad_x1_carbon_6.bin'
[CHIPSEC] Found SPI Flash descriptor at offset 0x0 in the binary 'lenovo_thinkpad_x1_carbon_6.bin'
Traceback (most recent call last):
  File "chipsec_util.py", line 232, in <module>
    sys.exit( main() )
  File "chipsec_util.py", line 228, in main
    return chipsecUtil.main()
  File "chipsec_util.py", line 195, in main
    comm.run()
[...]

PR #604, opened 2019-06-21
fixup evsa guid

BrentHoltsclaw:
address issue #603

### DIFF
--- a/chipsec/hal/uefi_platform.py
+++ b/chipsec/hal/uefi_platform.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2015, Intel Corporation
+#Copyright (c) 2010-2019, Intel Corporation
 # 
 #This program is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


chipsec/hal/uefi_platform.py
```diff
@@ -796,7 +796,7 @@ def EFIvar_EVSA(nvram_buf):
     sn = 0
     EVSA_RECORD = "<IIII"
     evsa_rec_size = struct.calcsize(EVSA_RECORD)
-    GUID_RECORD = "<HIHH8s"
+    GUID_RECORD = "<H4s2s2s8s"
```
npmitche: Why <H4s2s2s8s and not <H4s2s2s2s6s? Later on you're just dicing the last variable up anyways:
def guid_str(guid0, guid1, guid2, guid3):
    guid = uuid.UUID(bytes=guid0[::-1] + guid1[::-1] + guid2[::-1] + guid3[:2] + guid3[-6:])

BrentHoltsclaw: I was initially trying to do minimal touching of the old code.   Your change makes sense, but I would actually like all guid to use uuid.UUID.

PR #613, opened 2019-07-01
Removed workaround for RwEverything over Hyper-V

elicn:
A workaround for RWE driver causing CHIPSEC over Hyper-V to report SMRR is not supported regardless of used driver.
Since CHIPSEC manual states that it no longer supprts RwEverything driver on Windows, that workaround should be removed.
If this workaround is still needed for some reason, better tell RWE driver from CHIPSEC driver, and act accordingly.


ErikBjorge:
Thanks for the patch.

PR #622, opened 2019-07-16
Add support to use python build_ext -i command on windows

BrentHoltsclaw:
(assuming VS and WDK are installed)

### DIFF
--- a/chipsec_tools/compression/build.cmd
+++ b/chipsec_tools/compression/build.cmd
@@ -16,27 +16,30 @@
 @REM #Contact information:
 @REM #chipsec@intel.com
 
+goto :main
+
+:set_buildvars
+@echo Building compression
+@echo off
+@REM # Need to set EDK_TOOLS_PATH and BASE_TOOLS_PATH to current directory
+set EDK_TOOLS_PATH=%CD%
+set BASE_TOOLS_PATH=%CD%
+
+@REM # Now call make and build the compression tools
+call nmake -c
+
+goto :end
+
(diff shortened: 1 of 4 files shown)


chipsec_tools/compression/build.cmd
```diff
@@ -16,27 +16,30 @@
 @REM #Contact information:
 @REM #chipsec@intel.com
 
+goto :main
+
+:set_buildvars
+@echo Building compression
+@echo off
+@REM # Need to set EDK_TOOLS_PATH and BASE_TOOLS_PATH to current directory and HOST_ARCH
```
NicholasArmour: Does HOST_ARCH still need to be set?

setup.py
```diff
@@ -139,6 +139,28 @@ def _build_darwin_compression(self):
         for exe in exes:
             self.copy_file(os.path.join(build_exe,"bin",exe),dst)
 
+    def _build_win_driver(self):
+        log.info("building the windows driver")
+        build_driver = os.path.join("drivers", "win7")
+        cur_dir = os.getcwd()
+        os.chdir(build_driver)
+        print(os.getcwd())
```
NicholasArmour: debug print? do we need this?

chipsec_tools/compression/build.cmd
BrentHoltsclaw: should be 32 bit to compile, will remove the comment

setup.py
BrentHoltsclaw: good catch removing

PR #624, opened 2019-07-16
Fix trailing whitespace on TianoCompress command call

0xfede7c8:


chipsec/helper/linux/helper.py
```diff
@@ -1135,7 +1135,7 @@ def decompress_file( self, CompressedFileName, OutputFileName, CompressionType )
         elif CompressionType == chipsec.defines.COMPRESSION_TYPE_TIANO:
             decode_str = TIANO + decode_str
         elif CompressionType == chipsec.defines.COMPRESSION_TYPE_UEFI:
-            decode_str = EFI + decode_str + "--uefi"
+            decode_str = EFI + decode_str + "--uefi "
```
ErikBjorge: The fix is good.

In the future we may want to refactor all of this code to maintain a list of parameters as subprocess.check_output can take either a sequence of arguments or a string.  Then we don't have to worry about this type of error again.

PR #649, opened 2019-08-01
Add common multi-bus routines

frinzell:
Adding some commonly used routines to read and write multiple instances of a register, augmenting multi-bus support.

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2019, Intel Corporation
+#Copyright (c) 2010-2020, Intel Corporation
 #
 #This program is free software; you can redistribute it and/or
 #modify it under the terms of the GNU General Public License
@@ -24,7 +24,7 @@
 # -------------------------------------------------------------------------------
 #
 # CHIPSEC: Platform Hardware Security Assessment Framework
-# (c) 2010-2019 Intel Corporation
+# (c) 2010-2020 Intel Corporation
 #
 # -------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


chipsec/chipset.py
```diff
@@ -839,6 +849,12 @@ def read_register(self, reg_name, cpu_thread=0, bus_index=0):
 
         return reg_value
 
+    def read_register_all(self, reg_name, cpu_thread=0):
+        values = []
+        for index in range( len(self.get_register_bus( reg_name )) ):
```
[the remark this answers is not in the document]
frinzell: Register instances are accessed by their Bus index.

chipsec/chipset.py
```diff
@@ -855,6 +865,15 @@ def read_register(self, reg_name, cpu_thread=0, bus_index=0):
 
         return reg_value
 
+    def read_register_all(self, reg_name, cpu_thread=0):
+        values = []
+        bus_data = self.get_register_bus( reg_name )
+        if bus_data is None:
+            return None
```
BrentHoltsclaw: Should return the result of self.read_register( reg_name, cpu_thread)

chipsec/chipset.py
```diff
@@ -855,6 +865,15 @@ def read_register(self, reg_name, cpu_thread=0, bus_index=0):
...
+        for index in range( len(bus_data) ):
```
BrentHoltsclaw: get_register_bus returns a list of the buses, should just loop through those instances.  The instances are returned in string, so you'll need to convert them to integer.

BrentHoltsclaw: @frinzell please let me know why range is being used instead of the list of bus entries

frinzell: `index` in `self.read_register()`  is index (0,1,2...) of the bus list not the actual bus number (0x1D, 0x1F,..)

BrentHoltsclaw:
@frinzell please rebase

PR #650, opened 2019-08-01
Extend register to use memory type within configuration files

BrentHoltsclaw:
Add ability for register to reference memory as a type
Modified memory xml attribute renamed type to function

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -63,6 +63,7 @@ class RegisterType:
     IOBAR     = 'iobar'
     MSGBUS    = 'msgbus'
     MM_MSGBUS = 'mm_msgbus'
+    MEMORY    = 'memory'
 
 
 ##################################################################################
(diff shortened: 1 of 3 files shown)


chipsec/chipset.py
```diff
@@ -782,18 +783,26 @@ def is_register_defined(self, reg_name):
...
+                    dev = self.Cfg.CONFIG_PCI[dev_name]
+                    reg_def['bus'] = dev['bus']
+                    reg_def['dev'] = dev['dev']
+                    reg_def['fun'] = dev['fun']
+                    if dev_name in self.Cfg.BUS:
+                        if bus_index < len(self.Cfg.BUS[dev_name]):
+                            reg_def['bus'] = self.Cfg.BUS[dev_name][bus_index]
+                        else:
+                            logger().error( "Bus index {:d} for '{}' not found.".format(bus_index, dev_name) )
+            elif reg_def["type"] == "memory":
```
frinzell: Does `<memory>` have a `device` attribute?  Was this for `<mmio>`?

chipsec/chipset.py
```diff
@@ -834,6 +843,11 @@ def read_register(self, reg_name, cpu_thread=0, bus_index=0):
             reg_value = self.msgbus.msgbus_reg_read( int(reg['port'],16), int(reg['offset'],16) )
         elif RegisterType.MM_MSGBUS == rtype:
             reg_value = self.msgbus.mm_msgbus_reg_read(int(reg['port'],16), int(reg['offset'],16))
+        elif RegisterType.MEMORY == rtype:
+            if reg['access'] == 'dram':
+                reg_value= self.mem.read_physical_mem(int(reg['address'],16), int(reg['size'],16))
+            elif reg['access'] == 'mmio':
+                reg_value = self.mmio.read_MMIO_reg(int(reg['address'],16), int(reg['offset']),int(reg['size'],16))
```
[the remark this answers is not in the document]
BrentHoltsclaw: good catch will make the change

chipsec/chipset.py
BrentHoltsclaw: this is for defining the register with a device type memory.  Using the common.xml for example:
now I can create a register based upon TPM as follows:
<register name ="TEST" type="memory" device="TPM" offset="0x0" size="4" desc="Test" />

PR #651, opened 2019-08-02
Memlock upd

frinzell:
- Added additional lock bit support check
- Some code cleanup

### DIFF
--- a/chipsec/modules/common/memlock.py
+++ b/chipsec/modules/common/memlock.py
@@ -1,6 +1,7 @@
 #CHIPSEC: Platform Security Assessment Framework
 #Copyright (c) 2018, Eclypsium, Inc.
-# 
+#Copyright (c) 2019, Intel Corporation
+#
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


chipsec/modules/common/memlock.py
```diff
@@ -45,14 +46,16 @@ class memlock(chipsec.module_common.BaseModule):
...
+        self.res = ModuleResult.NOTAPPLICABLE
 
     def is_supported(self):
         # Workaround for Atom based processors.  Accessing this MSR on these systems
         # causes a GP fault and can't be caught in UEFI Shell.
-        if self.cs.get_chipset_id() in chipsec.chipset.CHIPSET_FAMILY_ATOM:
-            return False
-        return True
-        
+        if self.cs.get_chipset_id() not in chipsec.chipset.CHIPSET_FAMILY_ATOM:
```
ErikBjorge: This could be reduced to a self.cs.is_atom() call.  Not required unless you really want to change it.

BrentHoltsclaw: I think the same can be done in multiple modules, going to merge this in and make a more general change

PR #659, opened 2019-08-15
Cleanup use is for None type compare

BrentHoltsclaw:


BrentHoltsclaw:
updated copyrights

PR #670, opened 2019-08-29
Fix build errors on Debian 10 & Ubuntu 18.04

citypw:
Move the "from __future__*" declaration into the beginning


BrentHoltsclaw:
@citypw thank you for the PR.  I believe that #663 also contains a fix for the issue.  The print function is actually not used within the file and it was just removed within that version.

citypw:
Cool, I'll wait for the #663 to be merged and closing the PR. Thanks.

PR #686, opened 2019-09-11
Cleanup imports for tools.secureboot.te

BrentHoltsclaw:


chipsec/modules/tools/secureboot/te.py
```diff
@@ -70,7 +71,7 @@ def dumpstr(s):
             print ("")
     print ("")
 
-#logger.VERBOSE = False
+#logger().VERBOSE = False
```
ErikBjorge: Why are we fixing a comment?  Lets get rid of the dead code.

BrentHoltsclaw: Removed

PR #726, opened 2019-10-24
Fix decode functionality

BrentHoltsclaw:
Fix error when cannot decode EFI_SECTION_USER_INTERFACE
Fix error in File size logic within NextFwFile functionality

### DIFF
--- a/chipsec/hal/spi_uefi.py
+++ b/chipsec/hal/spi_uefi.py
@@ -263,7 +263,10 @@ def build_efi_modules_tree( _uefi, fwtype, data, Size, offset, polarity ):
                 sec.calc_hashes( sec.HeaderSize )
             elif sec.Type == EFI_SECTION_USER_INTERFACE:
                 # "leaf" UI section: update section's UI name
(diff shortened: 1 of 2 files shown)


chipsec/hal/spi_uefi.py
```diff
@@ -263,7 +263,10 @@ def build_efi_modules_tree( _uefi, fwtype, data, Size, offset, polarity ):
                 sec.calc_hashes( sec.HeaderSize )
             elif sec.Type == EFI_SECTION_USER_INTERFACE:
                 # "leaf" UI section: update section's UI name
-                sec.ui_string = sec.Image[sec.HeaderSize:-2].decode("utf-16")
+                try:
+                    sec.ui_string = sec.Image[sec.HeaderSize:-2].decode("utf-16")
+                except UnicodeDecodeError:
+                    pass
```
npmitche: Should you set the ui_string to something like "unknown" if it can't read it from the image?

BrentHoltsclaw: I don't know that a blank ui_string vs an unknown ui_string would be helpful.  Can make changes if others have consensus that it needs to be changed

PR #732, opened 2019-10-29
valid_mmap_phys_addr_range conflicts with upstream

sempervictus:
PaX implements valid_mmap_phys_addr_range in the kernel, and the
(return 1) hack in chipsec_km.c has a name collision with it.

Add another conditional for exposing the definition predicated on
whether __HAVE_ARCH_PAX_OPEN_USERLAND is defined in the kernel
headers. If it is, dont redefine the original function.

### DIFF
--- a/drivers/linux/chipsec_km.c
+++ b/drivers/linux/chipsec_km.c
@@ -547,7 +547,7 @@ int __weak phys_mem_access_prot_allowed(struct file *file,
         return 1;
 }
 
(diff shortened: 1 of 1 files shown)


sempervictus:
This hasn't been loaded into the kernel yet, just build-tested to not fail on the name collision.

BrentHoltsclaw:
@sempervictus thanks for the pull request

Issue #730, opened 2019-10-28
Possible mistake in definition for PWRMBASE for 300 series on-package PCH


martingalvan:
Hi, I'm looking at pch_3xx.xml and saw the following definition for `PWRMBASE`:

```
    <register name="PWRMBASE" type="pcicfg" bus="0" dev="0x1f" fun="2" offset="0x10" size="8" desc="PM Base Address">
      <field name="STYPE" bit="0"  size="1"  desc="Space Type (always 0 - memory space)"/>
      <field name="BA"    bit="13" size="51" desc="Base Address"/>
    </register>
```

The Intel 300 and C240 series PCH datasheet, vol 2 shows that there are two registers called `PWRMBASE` and `PWRMBASE HIGH`. I understand that the idea here is to merge them both into one, and the `BA` field will span from bit 13 of `PWRMBASE` to bit 31 of `PWRMBASE HIGH`. This works fine for "regular" 300 series PCHs; however, it appears that there's also an "on-package" 300 series PCH with its own datasheet ([here](https://www.intel.com/content/www/us/en/products/docs/chipsets/300-series-chipset-on-package-pch-datasheet-vol-2.html)). According to this datasheet, the base address field of `PWRMBASE` actually starts at bit 12, and bit 31 is reserved.

The reason I'm reporting this is that chipset.py lists this on-package PCH (DID 0x9D84) as part of the regular 300 series, which means that pch_3xx.xml will get picked for it. I think there should be specific xmls for these on-package PCHs.

BrentHoltsclaw:
Good find.  It appears that a pch_3xx-lp.xml will need to be created with the appropriate registers.

Sae86:
This issue (and mentioned issue) seem to have been resolved with #853, closing as fixed but please reopen if needed

Issue #729, opened 2019-10-28
Wrong definition for ACPI Base Address register in 300 series PCH?


martingalvan:
Hi, I'm looking at pch_3xx.xml and saw the following:

    <register name="ABASE" type="pcicfg" bus="0" dev="0x1f" fun="2" offset="0x40" size="4" desc="ACPI Base Address">
      <field name="STYPE" bit="0" size="1" desc="Space Type (always 1 - I/O space)"/>
      <field name="BA"    bit="8" size="8" desc="Base Address"/>
    </register>

However, the Intel 300 and C240 series PCH datasheet, vol 2 shows a different definition for this register (which it calls `ACPIBA`). Here it's described as a MSG Register at offset 0x27B4, with a 14-bit wide `ADDR` field starting from bit 2. The "On-Package" PCH datasheet shows the same.

Is this the same register? If so, why are the definitions different?

kc8apf:
As far as I know, my Whiskey Lake system has a 300 series PCH.  I can test an updated configuration.

BrentHoltsclaw:
Ok didn't understand that 3xx was broken.  Given the differences within the datasheets, it does not appear that a new configuration file needs to be created at this time.  Could you check #736.

martingalvan:
Well, the only relevant difference I saw between the on-package and regular versions is that of https://github.com/chipsec/chipsec/issues/730. I'll check #736 now, thanks!

kc8apf:
My Whiskey Lake system reports as follows:
> [CHIPSEC] API mode: using CHIPSEC kernel module API
> [CHIPSEC] OS      : Windows 10 10.0.19008 AMD64
> [CHIPSEC] Python  : 2.7.17 (64-bit)
> [CHIPSEC] Platform: Mobile 8th Generation Core Processor (Whiskey Lake U 4 Cores)
> [CHIPSEC]      VID: 8086
> [CHIPSEC]      DID: 3E34
> [CHIPSEC]      RID: 0B
> [CHIPSEC] PCH     : Intel 300 series On-Package PCH
> [CHIPSEC]      VID: 8086
> [CHIPSEC]      DID: 9D84
> [CHIPSEC]      RID: 30

I'm willing to run tests if someone describes what to run.

BrentHoltsclaw:
@kc8apf With the current version of chipsec and with #736.  If you could run `chipsec_util.py io list` and `chipsec_util.py mmio list`.

kc8apf:
```
[...]

PR #736, opened 2019-10-30
Changes to 3xx.xml to match datasheets

BrentHoltsclaw:
ABASE and PMBASE changed
PWRMBASE moved to help match 3xx-OP datasheets
Reference 337868-002 and 337348-001
#729 #730 

### DIFF
--- a/chipsec/cfg/pch_3xx.xml
+++ b/chipsec/cfg/pch_3xx.xml
@@ -33,7 +33,7 @@ XML configuration file for the 300 series PCH
   <!-- #################################### -->
   <io>
     <bar name="ABASE"      register="ABASE"    base_field="BA"    size="0x100" fixed_address="0x1800" desc="ACPI Base Address"/>
-    <bar name="PMBASE"     register="ABASE"    base_field="BA"    size="0x100" fixed_address="0x1800" desc="ACPI Base Address"/>
+    <bar name="PMBASE"     register="PMBASE"    base_field="BA"    size="0x100" fixed_address="0x1800" desc="ACPI Base Address"/>
     <bar name="TCOBASE"    register="TCOBASE"  base_field="TCOBA" size="0x20"  desc="TCO Base Address"/>
 
     <!-- old definition -->
@@ -60,8 +60,8 @@ XML configuration file for the 300 series PCH
     <!-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ -->
 
     <!-- Sideband Register Access Registers -->
-    <register name="SBREG_BAR" type="pcicfg" bus="0" dev="0x1F" fun="1" offset="0x10" size="4" desc="Sideband Register Access BAR">
-      <field name="RBA" bit="24" size="8" desc="Register Base Address"/>
+    <register name="SBREG_BAR" type="pcicfg" bus="0" dev="0x1F" fun="1" offset="0x10" size="8" desc="Sideband Register Access BAR">
+      <field name="RBA" bit="24" size="40" desc="Register Base Address"/>
     </register>
     <register name="P2SBC" type="pcicfg" bus="0" dev="0x1F" fun="1" offset="0xE0" size="2" desc="P2SB Configuration Register">
       <field name="HIDE" bit="8" size="1" desc="Hide SBREG_BAR"/>
@@ -71,13 +71,20 @@ XML configuration file for the 300 series PCH
     </register>
 
     <!-- Power Management Controller -->
-    <register name="ABASE" type="pcicfg" bus="0" dev="0x1f" fun="2" offset="0x40" size="4" desc="ACPI Base Address">
-      <field name="STYPE" bit="0" size="1" desc="Space Type (always 1 - I/O space)"/>
-      <field name="BA"    bit="8" size="8" desc="Base Address"/>
+     <register name="ABASE" type="mm_msgbus" port="0x88" offset="0x27B4" size="4" desc="ACPI Base Address">
+      <field name="MASK" bit="18" size="6" desc="MASK for lower 6 bits"/>
+      <field name="BA"    bit="2" size="14" desc="Base Address"/>
+    </register>
+    <register name="PMBASE" type="mm_msgbus" port="0x88" offset="0x27AC" size="4" desc="PM Base Address">
+      <field name="BA" bit="0" size="16" desc="Base Address" />
+      <field name="BAMRL" bit="16" size="16" desc="Memory Range Limit"/>
     </register>
     <register name="PWRMBASE" type="pcicfg" bus="0" dev="0x1f" fun="2" offset="0x10" size="8" desc="PM Base Address">
       <field name="STYPE" bit="0"  size="1"  desc="Space Type (always 0 - memory space)"/>
-      <field name="BA"    bit="13" size="51" desc="Base Address"/>
+      <field name="BA"    bit="12" size="52" desc="Base Address"/>
+      <!-- bit 31 on 3xx-OP is reserved and should be set to 0
+           BA starts at 12 on 3xx-OP and 13 on because 12 will be 0 on 3xx moving the
+           starting address to 12 -->
     </register>
     <register name="GEN_PMCON_1" type="mmio" bar="PWRMBASE" offset="0x1020" size="4" desc="General PM Configuration A">
     </register>


chipsec/cfg/pch_3xx.xml
```diff
@@ -60,8 +60,8 @@ XML configuration file for the 300 series PCH
     <!-- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~ -->
 
     <!-- Sideband Register Access Registers -->
-    <register name="SBREG_BAR" type="pcicfg" bus="0" dev="0x1F" fun="1" offset="0x10" size="4" desc="Sideband Register Access BAR">
-      <field name="RBA" bit="24" size="8" desc="Register Base Address"/>
+    <register name="SBREG_BAR" type="pcicfg" bus="0" dev="0x1F" fun="1" offset="0x10" size="8" desc="Sideband Register Access BAR">
```
martingalvan: I'd add a comment here indicating that this also includes SBREG_BARH, and that this should be done for 1xx.xml and 2xx.xml as well.

chipsec/cfg/pch_3xx.xml
```diff
@@ -71,13 +71,20 @@ XML configuration file for the 300 series PCH
     </register>
 
     <!-- Power Management Controller -->
-    <register name="ABASE" type="pcicfg" bus="0" dev="0x1f" fun="2" offset="0x40" size="4" desc="ACPI Base Address">
-      <field name="STYPE" bit="0" size="1" desc="Space Type (always 1 - I/O space)"/>
-      <field name="BA"    bit="8" size="8" desc="Base Address"/>
+     <register name="ABASE" type="mm_msgbus" port="0x88" offset="0x27B4" size="4" desc="ACPI Base Address">
+      <field name="MASK" bit="18" size="6" desc="MASK for lower 6 bits"/>
+      <field name="BA"    bit="2" size="14" desc="Base Address"/>
```
martingalvan: This is missing the ACPIBADE field (bit 0, size 1), though I don't think it'll be used for now.

chipsec/cfg/pch_3xx.xml
```diff
@@ -71,13 +71,20 @@ XML configuration file for the 300 series PCH
...
+    </register>
+    <register name="PMBASE" type="mm_msgbus" port="0x88" offset="0x27AC" size="4" desc="PM Base Address">
+      <field name="BA" bit="0" size="16" desc="Base Address" />
+      <field name="BAMRL" bit="16" size="16" desc="Memory Range Limit"/>
     </register>
     <register name="PWRMBASE" type="pcicfg" bus="0" dev="0x1f" fun="2" offset="0x10" size="8" desc="PM Base Address">
       <field name="STYPE" bit="0"  size="1"  desc="Space Type (always 0 - memory space)"/>
-      <field name="BA"    bit="13" size="51" desc="Base Address"/>
+      <field name="BA"    bit="12" size="52" desc="Base Address"/>
```
martingalvan: I think this is wrong. The definition was fine as it was for 3xx, it's 3xx-OP that has `BA` starting at bit 12. I'd keep this definition as it was, and add the new one to 3xx-op.xml.

chipsec/cfg/pch_3xx.xml
```diff
@@ -71,13 +71,20 @@ XML configuration file for the 300 series PCH
...
+      <!-- bit 31 on 3xx-OP is reserved and should be set to 0
+           BA starts at 12 on 3xx-OP and 13 on because 12 will be 0 on 3xx moving the
```
martingalvan: `and 13 on because...` -> `and 13 on 3xx because...`

chipsec/cfg/pch_3xx.xml
```diff
@@ -71,13 +71,20 @@ XML configuration file for the 300 series PCH
...
+           starting address to 12 -->
     </register>
     <register name="GEN_PMCON_1" type="mmio" bar="PWRMBASE" offset="0x1020" size="4" desc="General PM Configuration A">
```
martingalvan: If I'm not mistaken, the `SMI_LOCK` field changes for both 3xx and 3xx-op. Here it's defined as bit 8, while in earlier PCHs it's bit 4. Could you verify this please?

martingalvan:
I think there's a mistake in the `I2C_EN` field of the `SMBUS_HCFG` register. The datasheet indicates that it's 1 bit, but `size` is set to 2.

ErikBjorge:
Yes, the I2C_EN bit is wrong in multiple configuration files.

It probably never caused a problem because it would have overlapped with the soft reset bit and we would have only written a 0 or 1 to the bit.  Needs to be fixed in any case in all impacted configuration files.

chipsec/cfg/pch_3xx.xml
BrentHoltsclaw: @ErikBjorge, do you have an opinion on whether we should make a whole new configuration for 3xxOP if the BA for PWRMBASE for 3xx starts at bit 13 (bit 12 will be 0) and the BA for 3xxOP for PWRMBASE starts at bit 12

chipsec/cfg/pch_3xx.xml
BrentHoltsclaw: @martingalvan would you like to make the PR for the change in 1xxx and 2xx?

martingalvan: Sure. I won't be able to do this immediately though, but I can definitely contribute this.

martingalvan:
Yes, I'll include the `I2C_EN` fix in my 1xx/2xx PR.

chipsec/cfg/pch_3xx.xml
martingalvan: Keep in mind there's also the issue of `PWRMBASE` having bit 31 marked as reserved. I think that'll be a nuisance when trying to read both `PWRMBASE` and `PWRMBASE_HIGH` as one large register.

ErikBjorge: In general if you make the base address field the larger of the two sizes it will work fine and you do not need to create a new configuration file.  This has to do with how BARs are implemented.  Low bits that are not valid will be hard wired to zero to indicate required alignment and size.

With that in mind we should be able to set the start bit to 12 and have it work for both configurations.

martingalvan: @ErikBjorge I assume that if we set the start bit to 12 and assume that it's hardwired to 0, we'll still have to modify all the offsets for 3xx to make up for it?

ErikBjorge: It should be handled in the change of the field value seen in the change here since we do not overlap another defined field in the register.

Also, when reading the BA field to get the MMIO base the value is masked and not shifted.  Take a look at read_register_field and the preserve_field_position parameter.  When we get the BA field in the MMIO HAL we set this value to True.

I hope I answered your question.  If not, please clarify the offsets you think need to be updated.

martingalvan: Right, I wasn't fully aware of how BARs work. I was under the impression that only the base address field mattered, and we didn't preserve the field position. That makes more sense now.

In any case, bit 31 always being set to reserved/zero is still concerning me. In fact, the whole definition for that register seems to be broken in the 300 On-Package datasheet (at least in DN 337868-002). My other concern is that the `TYPE` field of this BAR is hardwired to zero, meaning that it's 32 bits wide, yet we have `PWRMBASE_HIGH` right after it, making the base address 64 bits wide.

ErikBjorge: Why should bit 31 being forced to 0 concern you?  The OS or any other use of the BAR would just read the value and mask out the lower bits (type and similar).  I have a feeling the reserved high bit is just to prevent overlap with other fixed MMIO registers near 4GB.

The type field should be checked by the PCI enumeration code during resource assignment and the BAR set based on that value (in this case below 4GB).  The fact that we have a potential 64bit address is not a problem.  In the future a version of the chip may support placing the region above 4GB.  The 64bit register definition just supports this flexibility.

martingalvan:
There's one more thing I found: for regular and On-Package 3xx, the PMC I/O registers are based off `BAR 2`, which is at dev 31, fun 2, offset 0x20. This includes `SMI_EN`, which we use for the `bios_smi` module. We should define both of these too.

BrentHoltsclaw:
@martingalvan please checkout #853

PR #737, opened 2019-10-30
Fix decoding of S3 bootscript EDKCompat opcodes

gch1p:
Decoded opcodes were not returned from the function.
Was broken since 2efeda434594041371cd0337f95737924f4e1db2. Bad merge, apparently.

Signed-off-by: Evgeny Zinoviev <me@ch1p.io>


gch1p:
s3 bootscript patching is also broken by the way. I'll look at it closer and send another PR.

PR #758, opened 2019-11-14
Fix unittest errors

BrentHoltsclaw:
Updated mock_helper to use BaseHelper
Updated util harness to correctly override helper with testhelper

### DIFF
--- a/tests/software/util.py
+++ b/tests/software/util.py
@@ -41,12 +41,10 @@ class TestChipsecUtil(unittest.TestCase):
         """
         fileno, self.log_file = tempfile.mkstemp()
         os.close(fileno)
-        oshelper._helper = None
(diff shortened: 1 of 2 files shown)


tests/software/util.py
```diff
@@ -58,8 +58,8 @@ def _chipsec_util(self, arg, helper_class=mock_helper.TestHelper):
         It verifies that no error is being reported. self.log will be populated
         with the output.
         """
-        oshelper._helper =  helper_class()
```
NicholasArmour: looks like oshelper._helper is no longer used.  should we remove it from setUp() and tearDown() as well?

BrentHoltsclaw: removed

PR #759, opened 2019-11-18
fix typo in duplicate did detection and update selecting data_dict if only 1 entry in the dict

NicholasArmour:
This fixes master not working

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -511,7 +511,7 @@ class Chipset:
                 self.did = 0xFFFF
                 self.rid = 0xFF
 
-        if self.did in Chipset_Dictionary and len(Chipset_Dictionary[self.did] > 1):
+        if self.did in Chipset_Dictionary and len(Chipset_Dictionary[self.did]) > 1:
             _unknown_platform = True
             self.longname = 'UnknownPlatform'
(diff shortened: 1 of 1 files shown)


chipsec/chipset.py
```diff
@@ -544,7 +544,7 @@ def init( self, platform_code, req_pch_code, start_driver, driver_exists=None, t
                 self.pch_rid = 0xFF
 
         if self.pch_vid == VID_INTEL and self.pch_did in pch_dictionary:
-            data_dict           = pch_dictionary[self.pch_did]
+            data_dict           = pch_dictionary[self.pch_did][0]
```
npmitche: Are you guaranteed to only have one element in each PCH?

NicholasArmour: Yes. The pch_dictionary definitions in chipset.py only have 1 list per pch_did key.

BrentHoltsclaw: At this time the answer is yes only 1.  We will need to update the logic if this changes.

npmitche: It might be good to check there's only one and if not, log a warning/info?

BrentHoltsclaw: @npmitche could you create a pull request or issue to add the logging

PR #760, opened 2019-11-18
Fix error messages and helper loading

BrentHoltsclaw:
Remove dependency for win32helper to find .sys file within init
Fixup logic to better display when driver file cannot be found

### DIFF
--- a/chipsec/helper/oshelper.py
+++ b/chipsec/helper/oshelper.py
@@ -91,7 +91,7 @@ class OsHelper:
         if(not self.helper):
             import platform
             os_system  = platform.system()
-            raise OsHelperError( "Could not load helper for '{}' environment (unsupported environment?)".format(os_system), errno.ENODEV )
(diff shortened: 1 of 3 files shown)


chipsec/helper/oshelper.py
```diff
@@ -91,7 +91,7 @@ def __init__(self):
         if(not self.helper):
             import platform
             os_system  = platform.system()
-            raise OsHelperError( "Could not load helper for '{}' environment (unsupported environment?)".format(os_system), errno.ENODEV )
+            raise OsHelperError( "Could not any load helpers for '{}' environment (unsupported environment?)".format(os_system), errno.ENODEV )
```
npmitche: "Could not any load helpers..."  Should be "Could not load any helpers..."?

BrentHoltsclaw: fixed

PR #768, opened 2019-11-25
utilcmd SMM communication protocol

NicholasArmour:
Added to HAL:
    - find_smmc: Scans physmem start-end looking for 'smmc'
    - send_smmc_SMI: prepares commbuffer, updates smmc structure, fires off SMI

Added to interrupts_cmd:
   - smmc switch that uses the RT_Code Start/End address, SMI GUID, Payload file to fire off SMI

### DIFF
--- a/chipsec/hal/interrupts.py
+++ b/chipsec/hal/interrupts.py
@@ -140,3 +140,36 @@ class Interrupts(hal_base.HALBase):
             #clear CommBuffer
             self.cs.helper.write_physical_mem(buf_addr,len(data_hdr),"\x00"*len(data_hdr))
             return None
+
+    # scan phys mem range start-end looking for 'smmc'
+    def scan_range(self,start,end):
+        chunk_sz = 1024 * 8 #8KB chunks
+        phys_address = start
+        found_at = 0
+        while phys_address <= end:
+            buffer = self.cs.mem.read_physical_mem( phys_address, chunk_sz )
+            buffer = chipsec.defines.bytestostring(buffer)
+            offset = buffer.find('smmc')
+            if offset != -1:
+                found_at = phys_address + offset
+                break
+            phys_address += chunk_sz
+        return found_at
+
+    def send_smmc_SMI(self,gSmmCorePrivate,guid,payload_fn,payload_loc):
+        guid_b = uuid.UUID(guid).bytes_le
+
+        payload_file = open('payload.bin','rb')
+        payload = payload_file.read()
+        payload_sz = len(payload)
+        payload_file.close()
+
+        data_hdr = guid_b + struct.pack("Q",payload_sz) + payload
+        # write payload to payload_loc
+        CommBuffer_offset = 56
+        BufferSize_offset = CommBuffer_offset + 8
+        self.cs.mem.write_physical_mem(gSmmCorePrivate + CommBuffer_offset,8,struct.pack("Q",payload_loc))
+        self.cs.mem.write_physical_mem(gSmmCorePrivate + BufferSize_offset,8,struct.pack("Q",payload_sz))
+        self.cs.mem.write_physical_mem(payload_loc,len(data_hdr),data_hdr)
+        self.send_SMI_APMC(0x0,0x0)
+
--- a/chipsec/hal/interrupts.py
+++ b/chipsec/hal/interrupts.py
@@ -142,7 +142,7 @@ class Interrupts(hal_base.HALBase):
             return None
 
     # scan phys mem range start-end looking for 'smmc'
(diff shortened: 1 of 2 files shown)


chipsec/utilcmd/interrupts_cmd.py
```diff
@@ -24,22 +24,29 @@
 
 from chipsec.command        import BaseCommand
 from chipsec.hal.interrupts import Interrupts
+import uuid
```
BrentHoltsclaw: remove from lgtm bot

chipsec/hal/interrupts.py
```diff
@@ -140,3 +140,53 @@ def send_ACPI_SMI(self, thread_id, smi_num, buf_addr, invoc_reg, guid, data):
...
+
+        data_hdr = guid_b + struct.pack("Q",payload_sz) + payload
+        # write payload to payload_loc
+        CommBuffer_offset = 56
+        BufferSize_offset = CommBuffer_offset + 8
+        self.cs.mem.write_physical_mem(gSmmCorePrivate + CommBuffer_offset,8,struct.pack("Q",payload_loc))
+        self.cs.mem.write_physical_mem(gSmmCorePrivate + BufferSize_offset,8,struct.pack("Q",payload_sz))
+        self.cs.mem.write_physical_mem(payload_loc,len(data_hdr),data_hdr)
+        self.send_SMI_APMC(0x0,0x0)
+
```
BrentHoltsclaw: Will a return value be put back in the buffer and is it worthwhile to read?

chipsec/utilcmd/interrupts_cmd.py
```diff
@@ -24,22 +24,29 @@
...
     >>> chipsec_util smi count
     >>> chipsec_util smi <thread_id> <SMI_code> <SMI_data> [RAX] [RBX] [RCX] [RDX] [RSI] [RDI]
+    >>> chipsec_util smi smmc <RT_code_start> <RT_code_end> <GUID> <payload_loc> <payload_file>
 
     Examples:
 
     >>> chipsec_util smi count
     >>> chipsec_util smi 0x0 0xDE 0x0
     >>> chipsec_util smi 0x0 0xDE 0x0 0xAAAAAAAAAAAAAAAA ..
+    >>> chipsec_util.py smi smmc 0x79dfe000 0x79efdfff ed32d533-99e6-4209-9cc02d72cdd998a7 0x79dfaaaa payload.bin
```
BrentHoltsclaw: Why does the payload need to be put in a file?  Could it also be passed in via cmd line argument?

chipsec/hal/interrupts.py
```diff
@@ -140,3 +140,53 @@ def send_ACPI_SMI(self, thread_id, smi_num, buf_addr, invoc_reg, guid, data):
             #clear CommBuffer
             self.cs.helper.write_physical_mem(buf_addr,len(data_hdr),"\x00"*len(data_hdr))
             return None
+
+    # scan phys mem range start-end looking for 'smmc'
+    def scan_range(self,start,end):
```
BrentHoltsclaw: should function name be more specific (find_smmc?)

chipsec/hal/interrupts.py
NicholasArmour: Yeah, there will be a return code... I should return that. Will add

chipsec/utilcmd/interrupts_cmd.py
NicholasArmour: I can make it string or file. Will add

p-state:
Where did you find using of such protocol? Isn't it the same as [send_ACPI_SMI](https://github.com/chipsec/chipsec/blob/4b2dd0dfc5a2f53e087fd54c896bb0961065d244/chipsec/hal/interrupts.py#L105)?

NicholasArmour:
The example I followed was how the variable runtime services [sends the commbuffer](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c#L242) using [mSmmCommunication->Communicate()](https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c#L471)

This accomplishes the same thing as send_ACPI_SMI, but that assumes your system publishes the UEFI ACPI table, and the system I was on did not do this.  [This commit](https://github.com/tianocore/edk2/commit/8b1d14939053b63d80355465649c50f9f391a64a#diff-8be2936b429473cb115b709422c51953) is where edk2 stopped publishing the ACPI table.

p-state:
@NicholasArmour wow, thank you

Issue #776, opened 2019-12-10
NUC DE3815TYBE .bio doesn't decompress properly


NicholasArmour:
TY0024.bio from https://downloadcenter.intel.com/download/29228/BIOS-Update-TYBYT20H-86A-

chipsec is missing a lot of files.

`uefi-firmware-parser -b TY0024.bio`  finds 232 pe's

NicholasArmour:
285 pe's found using PR#777

PR #777, opened 2019-12-11
Fixup spi_uefi File/Section parsing

BrentHoltsclaw:
#776 


BrentHoltsclaw:
@johnloucaides can you help validate the pull request?

0xfede7c8:
I will test with a bunch of images from different vendors and report back.

I will also propose the creation of a test suite for the UEFI parser, where we can store different images and the expected results. This way we can test changes faster in the future. I will create an issue for this and implement it if the schedule helps.

BrentHoltsclaw:
@0xfede7c8 any concern with this Pull Request?

0xfede7c8:
I have performed the manual test and didn't find anything bad. In fact I saw some extraction improvements for some images (more executables found). And I feel a small speed improvement, also.

0xfede7c8:
I have found this one: 
https://github.com/chipsec/chipsec/issues/792

I don't know if this pull request is intended to fix this issue but this bug seems to be present on both current master as well as this branch.

PR #779, opened 2019-12-11
Fix input buffer length check for IOCTL_RDMSR.

assafcarlsbad:
The check for the buffer length should reflect the fact that the input parameters for `IOCTL_RDMSR` are two 32-bit values (the new CPU thread id and the MSR address).


assafcarlsbad:
Force-pushed a fix to the trace message as well.

PR #782, opened 2019-12-12
Print DEVICE_OBJECT and DRIVER_OBJECT addresses using %p.

assafcarlsbad:
Printing addresses using the %x format specifier leads to truncated output on x64 platforms, for example:
image


assafcarlsbad:
Pushed some more commits related to the Windows driver with the aim of reducing the number of IntelliSense warnings encountered during the build. @BrentHoltsclaw @npmitche can you please re-review?

PR #796, opened 2019-12-20
Clean up spi_access and add check for valid flash descriptor

BrentHoltsclaw:


chipsec/modules/common/spi_access.py
```diff
@@ -1,6 +1,6 @@
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2017, Intel Corporation
-# 
+#Copyright (c) 2010-2019, Intel Corporation
```
NicholasArmour: copyright :)

BrentHoltsclaw: changes were made in 2019

PR #797, opened 2019-12-22
Fix RWEverything helper.

assafcarlsbad:
While support for the RWEverything helper is only marked as experimental, I find it convenient to use it from time to time since it saves the hassle of building the Windows driver and enabling test signing. This PR attempts to fix some issues with the RWEveything helper with the aim of making it a bit more functional.


BrentHoltsclaw:
@assafcarlsbad agree that we should keep this helper up to date as it can be helpful.  Be careful when using it though, I believe that @ErikBjorge deprecated its use as some of the results were coming back inconsistent (such as PCI read/write).

PR #822, opened 2020-01-13
Chipsec XML to Human-readable HTML log parser

mtnmts:
This is a log parser we have been using internally for easily going over chipsec's generated XML's, it generates a human-readable HTML file, with color-highlighting for each test case based on its status.

I work with @assafcarlsbad and this is my first contribution to Chipsec, Please let me know if I put things in a wrong directory, I couldn't find anywhere that looked suitable so I created_ `/utils`, if you want the parser but somewhere else just let me know and I'll fix the commit :).

it is a lxml parser and a flask/jinja2 server, after running
the server simply drag and drop the XML report and it will generate
a HTML report

Also included is an actual chipsec result XML for testing.

Thanks!


BrentHoltsclaw:
Sorry I'm a little backlogged, will review by the end of the week.

utils/python_parser/templates/template.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html>
+	<head>
+		<!-- Required meta tags -->
+		<meta charset="utf-8">
+		<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+
+		<!-- Bootstrap CSS -->
+		<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css" integrity="sha384-Vkoo8x4CGsO3+Hhxv8T/Q5PaXtkKtu6ug5TOeNV6gBiFeWPGFN9MuhOf23Q9Ifjh" crossorigin="anonymous">
```
BrentHoltsclaw: Is there an offline css template that can be loaded?

BrentHoltsclaw:
Do we want to create a new directory of utils or should this go under chipsec_tools?  I would lean more towards going within chipsec_tools anyone else have an opinion?

utils/python_parser/templates/template.html
mtnmts: Yes, Changed.

mtnmts:
@BrentHoltsclaw - I've changed to a offline CSS template & moved the directory to chipsec_tools (and renamed it to log_parser)

PR #823, opened 2020-01-14
Update the py2exe build scripts to be Python 3 compatible.

assafcarlsbad:
Afterwards, one can use an up-to-date fork of `py2exe` such as https://github.com/albertosottile/py2exe to build the entire CHIPSEC distribution using Python 3.

image

Building with Python 2 still works, of course.

### DIFF
--- a/scripts/build_exe_win7-amd64.py
+++ b/scripts/build_exe_win7-amd64.py
@@ -60,6 +60,7 @@ bin_dir = os.path.join(root_dir,"bin")
 source_dir = os.path.join(root_dir,"source")
 tool_dir   = root_dir
 cfg_dir    = os.path.join(tool_dir,"chipsec","cfg")
+version_file = os.path.join(root_dir, "chipsec", VERSION_FILE)
 
 win_7_amd64 = os.path.join(bin_dir,'win7-amd64');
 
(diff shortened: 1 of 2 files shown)


scripts/build_exe_win7-amd64.py
```diff
@@ -23,8 +23,8 @@
 
 # -------------------------------------------------------------------------------
 #
-# CHIPSEC: Platform Hardware Security Assessment Framework
-# (c) 2010-2012 Intel Corporation
+# CHIPSEC: Platform Security Assessment Framework
+# (c) 2010-2020 Intel Corporation
```
BrentHoltsclaw: Only Intel employees should be updating the copyright year for Intel.  You can either add your own/company copyright or waive the rights to add to the copyright.

scripts/build_exe_win7-x86.py
```diff
@@ -23,8 +23,8 @@
 
 # -------------------------------------------------------------------------------
 #
-# CHIPSEC: Platform Hardware Security Assessment Framework
-# (c) 2010-2012 Intel Corporation
+# CHIPSEC: Platform Security Assessment Framework
+# (c) 2010-2020 Intel Corporation
```
BrentHoltsclaw: see above comment

PR #829, opened 2020-01-16
Check that 'uefispec_concern' is not empty before logging an error.

assafcarlsbad:
CHIPSEC currently logs an error message even if no variables with attributes different from the UEFI spec were encountered:
image

### DIFF
--- a/chipsec/modules/common/uefi/access_uefispec.py
+++ b/chipsec/modules/common/uefi/access_uefispec.py
@@ -1,5 +1,5 @@
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2015, Intel Corporation
+#Copyright (c) 2010-2020, Intel Corporation
 # 
 #This program is free software; you can redistribute it and/or
 #modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


chipsec/modules/common/uefi/access_uefispec.py
```diff
@@ -178,10 +178,11 @@ def check_vars(self, do_modify):
                             self.logger.log_bad("Variable {} should be read only.".format(name))
                             res = ModuleResult.FAILED
 
-        self.logger.log('')
-        self.logger.log_bad('Variables with attributes that differ from UEFI spec:')
-        for name in uefispec_concern:
-            self.logger.log('    {}'.format(name))
+        if len(uefispec_concern) > 0:
```
BrentHoltsclaw: would it be better to use:
`if uefispec_concern:`

frinzell:
Looks good, just copyright year.

chipsec/modules/common/uefi/access_uefispec.py
assafcarlsbad: 👍

PR #839, opened 2020-01-24
Print updates

iherna2:
Did some refactoring of the print_buffer routines in the logger. 
Introduced a change to include the exit string in the result file, which is useful when exporting to json or xml. 
Added a new markdown output format, the markdown reports look nice on VSCode

### DIFF
--- a/chipsec/module.py
+++ b/chipsec/module.py
@@ -1,5 +1,5 @@
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2019, Intel Corporation
+#Copyright (c) 2010-2020, Intel Corporation
 #
 #This program is free software; you can redistribute it and/or
 #modify it under the terms of the GNU General Public License
@@ -41,6 +41,17 @@ class Module():
         self.module = None
         self.mod_obj = None
 
+    def __lt__(self, other):
+        return self.name < other.name
+
+    def __le__(self, other):
+        return self.name <= other.name
+    
+    def __gt__(self, other):
+        return self.name > other.name
+
+    def __ge__(self, other):
+        return self.name >= other.name
 
     def get_name(self):
         return self.name
--- a/chipsec/testcase.py
+++ b/chipsec/testcase.py
@@ -1,5 +1,5 @@
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2018-2019, Intel Corporation
+#Copyright (c) 2018-2020, Intel Corporation
 #
 #This program is free software; you can redistribute it and/or
 #modify it under the terms of the GNU General Public License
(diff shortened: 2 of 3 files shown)


chipsec/logger.py
```diff
@@ -208,10 +208,17 @@ def flush(self):
...
         if self.Results.get_current() is not None:
+            if hasattr(text, '__getitem__'):
+                text = "".join(text)
+            elif not text.endswith('\n'):
+                text += ('\n')
             self.Results.get_current().add_output(text)
+
+    def log( self, text, level=pyLogging.INFO):
+        """Sends plain text to logging."""
+        self.add_text_to_results(text)
```
BrentHoltsclaw: It looks like log calls add_text_to_results, why is this added to each logging function that also calls log?

chipsec/logger.py
```diff
@@ -381,87 +401,27 @@ def logger():
...
-        i+=1
-
-    if 0 != len(arr)%length:
-        tmp+=[ (length - len(arr)%length) * 3*" " ]
-        tmp+=["| "]
-        tmp+=tmp_str
-        tmp_s = "".join(tmp)
-        logger().log( tmp_s )
-
+    res = dump_buffer(arr, length)
```
BrentHoltsclaw: The changes are causing python2 to break.  For example bios_kbrd_buffer.  It is also breaking the util commands, for example chipsec_util.py acpi table UEFI

chipsec/logger.py
iherna2: The idea was to have whatever is being logged by the tests show up as well in the results. Currently if you export the results to xml, json (or the proposed markdown) you don't get to see in the file what is the test complaining about when it is failing.

chipsec/logger.py
iherna2: bummer.  Will need to find another way to achieve the same being compatible

chipsec/logger.py
BrentHoltsclaw: I am saying the calls on line 233, 277, 283, etc are redundant as each function calls log where your function is called.  I think that it can be condensed to only calling it within the log function

chipsec/logger.py
```diff
@@ -381,87 +401,27 @@ def logger():
...
-    if 0 != len(arr)%length:
-        tmp+=[ (length - len(arr)%length) * 3*" " ]
-        tmp+=["| "]
-        tmp+=tmp_str
-        tmp+=["\n"]
-    return "".join(tmp)
+
+    result = ""
+    pieces = chunk(arr, length)
+    for piece in pieces:
```
BrentHoltsclaw: Try this:
```
if isinstance(arr,str):
    left = "".join(["{0>2x} ".format(ord(c)) for c in piece])
    right = "".join("{["{:s} ".format(c) if c in string.printable else '. ' for c in piece ])
    line = "{: <{}} | {:s}\n".format(left,length*3,right)
else:
    line 413 - 415
```

chipsec/logger.py
iherna2: I see what you mean. The way I can see we can simplify this change would be to make a new class deriving from the logging module and then make all the callers use this new class.

It is cleaner and easier to maintain but a little more invasive. I can work on that idea if that sounds good to you.

chipsec/logger.py
iherna2: This can work for time being. I started exploring something based on bytearray that works for the bios_kbrd_buffer case on both Py2 and Py3.

but then started looking into the acpi/UEFI failure and then I found that there is a lot of code doing encode/decode operations to read binary blobs and print them out to screen. I think we probably should think refactoring and unify the code in a compatible way. I'll post a follow up based on your proposal.

iherna2: This is already taken care of in #878, no need to do anymore

iherna2:
Dropped some commits that were better done in #878

PR #841, opened 2020-01-30
Reputation checks for EFI modules based on the VirusTotal API.

assafcarlsbad:
Currently CHIPSEC is capable of conducting a forensic analysis of a firmware image either by comparing individual modules against a blacklist or a whitelist. This PR attempts to augment these capabilities by leveraging the VirusTotal API to perform reputation checks on individual EFI modules as they are read from the firmware dump.

### DIFF
--- a/chipsec/modules/tools/uefi/reputation.py
+++ b/chipsec/modules/tools/uefi/reputation.py
@@ -1,0 +1,133 @@
+"""
+This module checks current contents of UEFI firmware ROM or specified firmware image for bad EFI binaries as per the
+VirusTotal API. These can be EFI firmware volumes, EFI executable binaries (PEI modules, DXE drivers..) or EFI sections.
+The module can find EFI binaries by their UI names, EFI GUIDs, MD5/SHA-1/SHA-256 hashes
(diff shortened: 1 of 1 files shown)


BrentHoltsclaw:
Should probably add some kind of installation instructions.  May also want to gracefully fail if unable to import virus_total_apis.

chipsec/modules/tools/uefi/reputation.py
```diff
@@ -0,0 +1,142 @@
...
+        self.image = None
+        self.vt_threshold = 10
+        self.vt = None
+
+    def is_supported(self):
+        if has_virus_total_apis:
+            return True
+        else:
+            self.logger.log_important("""Can't import module 'virus_total_apis'.
+Please run 'pip install virus_total_apis' and try again.""")
```
npmitche: Per [pypi](https://pypi.org/project/virustotal-api/), it should be:
```bash
pip install virustotal-api
```

assafcarlsbad: My bad, fixed.

PR #843, opened 2020-02-01
Fix decompress pfs and readability

BrentHoltsclaw:

### DIFF
--- a/chipsec/hal/spi_uefi.py
+++ b/chipsec/hal/spi_uefi.py
@@ -176,7 +176,7 @@ class EFI_MODULE(object):
         if self.MD5   : _s  = "\n{}MD5   : {}".format(_ind,self.MD5)
         if self.SHA1  : _s += "\n{}SHA1  : {}".format(_ind,self.SHA1)
         if self.SHA256: _s += "\n{}SHA256: {}".format(_ind,self.SHA256)
-        return _s
+        return bytestostring(_s)
 
     def calc_hashes( self, off=0 ):
         if self.Image is None: return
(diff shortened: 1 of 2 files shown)


0xfede7c8:
I'm still getting differences between parsing UEFI files with python3 and python2.

Check attached file:

`python ./chipsec_main.py -n -m tools.uefi.whitelist -a generate,hashes.json,6AC10`
`python3 ./chipsec_main.py -n -m tools.uefi.whitelist -a generate,hashes.json,6AC10`

[6AC10.zip](https://github.com/chipsec/chipsec/files/4148183/6AC10.zip)

BrentHoltsclaw:
It looks like the difference in hash is due to the calculation using  bytes (python3) and string (python2).

0xfede7c8:
A few questions:

> It looks like the difference in hash is due to the calculation using bytes (python3) and string (python2).

You mean that this patch should fix the problem? 

Is the idea to support two languages in the same codebase? It feels a bit weird to me. At least have two different branches seems a good idea.

BrentHoltsclaw:
This patch does not fix the issue.  Python 2 and 3 were taking different decompression paths because python3 did not understand pfs.  This patch is to fix that piece.

BrentHoltsclaw:
The plan is to drop support for Python 2 after June 2020 until then we need to maintain 2/3 compatibility

0xfede7c8:
Is this going to be merged? I understood that this patch shouldn't fix the python2 python3 issues so I don't see any other problems.

PR #844, opened 2020-02-01
Common.py modifications

BrentHoltsclaw:
Remove duplicate configuration from common.py 
Move specific configuration to hal files


chipsec/hal/spi.py
```diff
@@ -202,26 +230,26 @@ def __init__(self, cs):
...
+    #def get_SPI_MMIO_base_fallback(self):
+    #    reg_value = self.cs.pci.read_dword( Cfg.SPI_MMIO_BUS, Cfg.SPI_MMIO_DEV, Cfg.SPI_MMIO_FUN, Cfg.SPI_MMIO_REG_OFFSET )
+    #    spi_base = ((reg_value >> Cfg.SPI_BASE_ADDR_SHIFT) << Cfg.SPI_BASE_ADDR_SHIFT) + Cfg.SPI_MMIO_BASE_OFFSET
+    #    if self.logger.HAL: self.logger.log( "[spi] SPI MMIO base: 0x{:016X} (assuming below 4GB)".format(spi_base) )
+    #    return spi_base
 
     def get_SPI_MMIO_base(self):
         if self.mmio.is_MMIO_BAR_defined('SPIBAR'):
             (spi_base,spi_size) = self.mmio.get_MMIO_BAR_base_address('SPIBAR')
-        else:
```
NicholasArmour, line 222: if SPIBAR is not defined, i believe this function will error out because spi_base will be undefined. should either initialize spi_base and spi_size, or keep the fallback code in there i think.

frinzell: **SPIBAR** is defined in **common.xml**, we may never hit the ELSE condition.  If it does, `get_SPI_MMIO_base_fallback()` looks like a hard coded version of the **SPIBAR** def in XML.

PR #869, opened 2020-03-12
Use Argparse in uefi_cmd.py

NicholasArmour:
My intent is to keep all functionality 100% the same as it was before it used argparse.

### DIFF
--- a/chipsec/utilcmd/uefi_cmd.py
+++ b/chipsec/utilcmd/uefi_cmd.py
@@ -37,7 +37,7 @@ from chipsec.hal.uefi          import *
 from chipsec.hal.spi_uefi      import *
 
 from chipsec.command    import BaseCommand
-
+from argparse import ArgumentParser
 
 # Unified Extensible Firmware Interface (UEFI)
 class UEFICommand(BaseCommand):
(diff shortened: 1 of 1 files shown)


chipsec/utilcmd/uefi_cmd.py
```diff
@@ -72,303 +72,367 @@ class UEFICommand(BaseCommand):
...
+        parser_assemble.add_argument('guid', type=str, help='guid')
+        parser_assemble.add_argument('file_type', type=str, help='')
+        parser_assemble.add_argument('comp', type=str, help='')
+        parser_assemble.add_argument('raw_file', type=str, help='')
+        parser_assemble.add_argument('efi_file', type=str, help='')
+        parser_assemble.set_defaults(func=self.assemble)
+
+        parser.parse_args(self.argv[2:], namespace=self)
+
         # No driver required when printing the util documentation
```
BrentHoltsclaw:
```suggestion
        # No driver for decode functionality
        if self.argv[2] in ('decode'):
            return False
        return True
```
Believe that 181 - 191 can be simplified to the above

chipsec/utilcmd/uefi_cmd.py
```diff
@@ -178,17 +178,11 @@ def requires_driver(self):
 
         parser.parse_args(self.argv[2:], namespace=self)
 
-        # No driver required when printing the util documentation
-        if len(self.argv) < 3:
+        # No driver for decode functionality
+        if self.argv[2] in ('decode'):
```
npmitche: Did you want in ['decode']? Currently if self.argv[2] = "d" this will return True.

PR #783, opened 2019-12-13
Change XML and Configuration Loading

BrentHoltsclaw:
#741

PR #871, opened 2020-03-17
Fixup configuration issues when multiple configuration files for same platform are loaded

BrentHoltsclaw:
Fix for #783.  Allows multiple configuration files for the same platform/pch to be loaded.
Also fixes selection of platform and pch


BrentHoltsclaw:
Merged changes into #783

PR #880, opened 2020-03-31
Misc changes for public

iherna2:
Miscellaneous changes

### DIFF
--- a/chipsec/helper/linux/linuxhelper.py
+++ b/chipsec/helper/linux/linuxhelper.py
@@ -1178,7 +1178,7 @@ class LinuxHelper(Helper):
     #
     def get_threads_count ( self ):
         import subprocess
-        return int(subprocess.check_output("grep -c process /proc/cpuinfo", shell=True))
+        return int(subprocess.check_output("egrep -c ^process /proc/cpuinfo", shell=True))
 
 def get_helper():
     return LinuxHelper()
--- a/chipsec/modules/common/cpu/cpu_info.py
+++ b/chipsec/modules/common/cpu/cpu_info.py
@@ -49,7 +49,7 @@ class cpu_info(BaseModule):
             self.logger.log('[*] Thread {:04d}'.format(thread))
 
             # Get processor brand string
-            brand = ''
+            brand = b''
             for eax_val in [0x80000002, 0x80000003, 0x80000004]:
                 regs = self.cs.cpu.cpuid(eax_val, 0)
                 for i in range(4):
--- a/chipsec/modules/common/rtclock.py
+++ b/chipsec/modules/common/rtclock.py
@@ -34,7 +34,10 @@ class rtclock(BaseModule):
         self.res = ModuleResult.PASSED
 
     def is_supported(self):
-        return self.cs.is_core()
+        if self.cs.is_core():
+            return True
+        self.res = ModuleResult.NOTAPPLICABLE
+        return False
 
     def check_rtclock(self):
         self.logger.start_test( "Protected RTC memory locations" )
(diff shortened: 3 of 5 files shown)


chipsec/helper/linux/linuxhelper.py
```diff
@@ -1178,7 +1178,7 @@ def decompress_file( self, CompressedFileName, OutputFileName, CompressionType )
     #
     def get_threads_count ( self ):
         import subprocess
-        return int(subprocess.check_output("grep -c process /proc/cpuinfo", shell=True))
+        return int(subprocess.check_output("egrep -c ^process /proc/cpuinfo", shell=True))
```
BrentHoltsclaw, line 1181: Why egrep vs grep?

chipsec/modules/common/cpu/cpu_info.py
```diff
@@ -49,7 +49,7 @@ def run(self, module_argv):
             self.logger.log('[*] Thread {:04d}'.format(thread))
 
             # Get processor brand string
-            brand = ''
+            brand = b''
```
BrentHoltsclaw, line 52: This seems to be breaking the module

chipsec/modules/common/rtclock.py
```diff
@@ -34,7 +34,10 @@ def __init__(self):
         self.res = ModuleResult.PASSED
 
     def is_supported(self):
-        return self.cs.is_core()
+        if self.cs.is_core():
+            return True
+        self.res = ModuleResult.NOTAPPLICABLE
+        return False
```
BrentHoltsclaw, line 40: Need to do some more reading, however this looks like it's not implemented for atom

chipsec/modules/memconfig.py
```diff
@@ -60,7 +60,10 @@ def __init__(self):
         BaseModule.__init__(self)
 
     def is_supported(self):
-        return self.cs.is_core()
+        if self.cs.is_core():
+            return True
+        self.res = ModuleResult.NOTAPPLICABLE
```
BrentHoltsclaw, line 65: Same as above.  Believe this is also not implmented for Atom.

chipsec/modules/smm_dma.py
```diff
@@ -43,6 +43,7 @@ def __init__(self):
         BaseModule.__init__(self)
 
     def is_supported(self):
+        self.res = ModuleResult.NOTAPPLICABLE
         if self.cs.is_atom(): return False
```
BrentHoltsclaw, line 47: Same as above.  Believe this is also not implemented for Atom

chipsec/helper/linux/linuxhelper.py
iherna2: not really needed as the regex is pretty simple. I somehow think we should use the info from MSRs rather than parsing a tool.

chipsec/modules/common/cpu/cpu_info.py
iherna2: for Python2?

chipsec/modules/common/rtclock.py
iherna2: This patch should not change the functionality, only the result. Do you prefer to find if is applicable to atom and make a single commit?

BrentHoltsclaw: Understand that functionality does not change.  The problem is the meaning of the results. 
 Not Applicable means that the test does not apply to a platform (registers don't exist or don't function in the same way).  

If you want to fix up the server piece that should be fine, but lets not categorize Atom until there's a definitive answer

PR #890, opened 2020-04-17
GLK: fix ABASE definition

NicholasArmour:
https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/silver-celeron-datasheet-vol-2.pdf  pg 958


BrentHoltsclaw:
@NicholasArmour please rebase

PR #892, opened 2020-04-20
linux helper HAS_EFI bugfix

Cr4sh:
Linux helper Makefile uses `grep -c 'D efi' /proc/kallsyms` search string in order to check if kernel image was built with EFI support which is quite unreliable, it is better to grep by `efi_call`.
Here's some example from my Gentoo system:
```
# uname -a
Linux HWSEC 4.19.97-gentoo-x86_64 #1 SMP Thu Mar 26 09:52:53 -00 2020 x86_64 Intel(R) Atom(TM) Processor E3950 @ 1.60GHz GenuineIntel GNU/Linux
# grep -c 'D efi' /proc/kallsyms
0
# grep -c 'efi_call' /proc/kallsyms
7
```

### DIFF
--- a/drivers/linux/Makefile
+++ b/drivers/linux/Makefile
@@ -24,7 +24,7 @@ check_kernel_dir:
 
 chipsec: check_kernel_dir clean
 	nasm -f $(elf-size) -o $(asm-path)/cpu.o $(asm-path)/cpu.asm
-	@if [ `grep -c 'D efi' /proc/kallsyms` != "0" ]; then \
+	@if [ `grep -c 'efi_call' /proc/kallsyms` != "0" ]; then \
 	    make CFLAGS_MODULE=-DHAS_EFI=1 -C $(KERNEL_SRC_DIR) M=${CURDIR} modules; \
 	else \
 	    make -C $(KERNEL_SRC_DIR) M=${CURDIR} modules; \


iherna2:
Is there a reason why we don't just test for CONFIG_EFI in chipsec_km.c ?
CONFIG_EFI comes from the kernel build system which should be more robust than testing the runtime state in the system we are building on and then defining our own variable.

Cr4sh:
@iherna2 I'm not really a Linux kernel programming expert so I suggested there's some particular reason for this sort of hack. But indeed, CONFIG_EFI approach looks more adequate.

BrentHoltsclaw:
@iherna2 When looking into the issue a while ago, some distributions were allowing CONFIG_EFI to be in both legacy and UEFI boot builds which was causing issues.  If this isn't the case anymore that is a better solution.

iherna2:
@BrentHoltsclaw I see what is the problem, and I don't think it will go away anytime soon, the distros kernel have EFI support compiled in always. 

This runtime test will work as long as there is a /proc/kallsyms readable, some kernels turn that readable only to root. That would mean building as root which should not be a major problem.

Maybe "test -f /sys/firmware/efi/fw_vendor" or something along those lines could work as well. I am ok  with the patch as posted

Issue #895, opened 2020-04-25
Apl.xml & Glk.xml SPIBAR wrong mask


Kelvinhack:
I doubt the correctness of SPIBAR valid bit mask in Apl.xml and Glk.xml, unlike other Intel Atom chipset , Apollo and Gemini Lake datasheet stated that the mask should be bit [31:12] (0xFFFFF000) , instead of bit[31:9] (0xFFFFFE00)

![image](https://user-images.githubusercontent.com/22551808/80280156-2e2c1700-86b7-11ea-98fd-0ae1bf90a33c.png)

![image](https://user-images.githubusercontent.com/22551808/80280171-4b60e580-86b7-11ea-80b9-f54769066b76.png)

Please correct me if I'm wrong.

BrentHoltsclaw:
@Kelvinhack yes the masks are incorrect.

frinzell:
Closing as fixed.

PR #899, opened 2020-04-29
Apollo Lake/Gemini Lake: Fix register definitions

martingalvan:
This commit fixes several register/field definitions for APL and GLK. I found
these mistakes after checking against the Intel datasheets.

In some cases (such as BDSM and friends) I saw that we can just fall back to the
'common' definitions, since those match the datasheets'.

Notice that I intentionally didn't fix ABASE for GLK, so as not to conflict with
PR #890.


BrentHoltsclaw:
@martingalvan any chance you can fixup the spibar from #895 within this PR?

PR #921, opened 2020-05-29
rtclock.py improvements

frinzell:
- Add modify switch and test
- Add AVN platform to is_supported()


chipsec/modules/common/rtclock.py
```diff
@@ -23,44 +23,83 @@
 """
 
 from chipsec.module_common import BaseModule, ModuleResult, MTAG_BIOS, MTAG_HWCONFIG
-
+from chipsec.hal.cmos import CMOS, CmosRuntimeError
```
BrentHoltsclaw, line 26: please fix from lgtm

frinzell: `CmosRuntimeError` removed

PR #936, opened 2020-06-19
Add cpuid option when identifing cpu

BrentHoltsclaw:
If cpuid is within configuration file, can assign cpu configuration
without did

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -393,7 +393,7 @@ class Chipset:
                             if _info.attrib['family'].lower() == "quark":
                                 CHIPSET_FAMILY_QUARK.append(_cfg.attrib['platform'].upper())
                         if 'detection_value' in _info.attrib:
-                            for dv in list(_info.attrib['detection_value'].split(', ')):
+                            for dv.strip() in list(_info.attrib['detection_value'].split(',')):
                                 self.detection_dictionary[dv] = _cfg.attrib['platform'].upper()
                         if _info.iter('sku'):
                             for _sku in _info.iter('sku'):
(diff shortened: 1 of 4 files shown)


chipsec/chipset.py
```diff
@@ -243,6 +243,12 @@ def init( self, platform_code, req_pch_code, start_driver, driver_exists=None, t
                 self.vid = vid
                 self.did = did
                 self.rid = rid
+            elif cpuid in self.detection_dictionary.keys():
+                self.code       = self.detection_dictionary[cpuid]
```
frinzell, line 247: Do we need `_unknown_platform = False` ?

chipsec/chipset.py
```diff
@@ -384,6 +391,9 @@ def init_xml_configuration(self):
                                 CHIPSET_FAMILY_XEON.append(_cfg.attrib['platform'].upper())
                             if _info.attrib['family'].lower() == "quark":
                                 CHIPSET_FAMILY_QUARK.append(_cfg.attrib['platform'].upper())
+                        if 'detection_value' in info.attrib:
```
frinzell, line 394: Should be `_info.attrib`

chipsec/chipset.py
BrentHoltsclaw: Thanks :)

chipsec/chipset.py
```diff
@@ -384,6 +392,9 @@ def init_xml_configuration(self):
                                 CHIPSET_FAMILY_XEON.append(_cfg.attrib['platform'].upper())
                             if _info.attrib['family'].lower() == "quark":
                                 CHIPSET_FAMILY_QUARK.append(_cfg.attrib['platform'].upper())
+                        if 'detection_value' in _info.attrib:
+                            for dv in list(_info.attrib['detection_value'].split(', ')):
```
frinzell, line 396: It might be better to strip spaces just in case.  How about this...  `for dv in [value.strip() for value in _info.attrib['detection_value'].split(',')]:`

BrentHoltsclaw: updated to  `for dv.strip() in list(_info.attrib['detection_value'].split(','))`.  Believe that they are equivalent.

PR #941, opened 2020-06-26
Update ia_untrusted

BrentHoltsclaw:
check soc_bios_done bit
check each thread for IA_Untrusted bit

### DIFF
--- a/chipsec/modules/common/cpu/ia_untrusted.py
+++ b/chipsec/modules/common/cpu/ia_untrusted.py
@@ -28,19 +28,42 @@ class ia_untrusted(BaseModule):
         BaseModule.__init__(self)
 
     def is_supported(self):
-        if self.cs.is_register_defined('MSR_BIOS_DONE') and self.cs.register_has_field('MSR_BIOS_DONE', 'IA_UNTRUSTED'):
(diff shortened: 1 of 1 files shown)


chipsec/modules/common/cpu/ia_untrusted.py
```diff
@@ -28,19 +28,42 @@ def __init__(self):
...
+        if self.cs.register_has_field('MSR_BIOS_DONE', 'SoC_BIOS_DONE'):
+            soc = self.cs.read_register_field('MSR_BIOS_DONE', 'SoC_BIOS_DONE')
+            if soc == 0:
+                res = ModuleResult.FAILED
+                self.logger.log_bad('SoC_BIOS_DONE not set.')
+            else:
+                self.logger.log_good('SoC_BIOS_DONE set.')
+
+        self.logger.log("")
+        for tid in range(self.cs.msr.get_cpu_thread_count()):
```
iherna2, line 49: We really should have a helper to verify a register is programmed the same way on every instance.

BrentHoltsclaw: Agree

PR #951, opened 2020-07-02
More whitespace cleanup

BrentHoltsclaw:
E231, E201, E226


npmitche:
Is there a rule that puts a space on the right side of operators (+, *, etc)? If so, could you add that to this list? It's putting spaces on the left, but would be nice to put both in the same PR.

Issue #953, opened 2020-07-05
Blackarch - DKMS make error


PaintedBlck:
Dear ChipSec development team,

since some time, chipsec produces an error when trying to install it with DKMS (or when doing a system update). Some help from your side would be welcome.
- Is there a manual way to build and install that can do the trick?
- I join some error information below (if you need more info, please let me know)

Kind regards,
PaintedBlck
 
_**The system**_:
- x64 hardware
- Linux BlackArch - 5.7.7-arch1-1

_**Method of installing**_:
- `git clone https://aur.archlinux.org/chipsec-git.git`
- `cd chipsec-git`
- `makepkg -Acs`
- `sudo pacman -U chipsec-git-1.5.2.r2.gaf7fa73-1-x86_64.pkg.tar.xz`

_**The error message**_:
```
==> dkms install --no-depmod -m chipsec -v 964.af7fa73 -k 5.7.7-arch1-1
Error! Bad return status for module build on kernel: 5.7.7-arch1-1 (x86_64)
Consult /var/lib/dkms/chipsec/964.af7fa73/build/make.log for more information.
==> Warning, `dkms install --no-depmod -m chipsec -v 964.af7fa73 -k 5.7.7-arch1-1' returned 10
==> depmod 5.7.7-arch1-1
```

_**The log file /var/lib/dkms/chipsec/964.af7fa73/build/make.log**_ :
```
DKMS make.log for chipsec-964.af7fa73 for kernel 5.7.7-arch1-1 (x86_64)
Sun 05 Jul 2020 03:09:03 PM CEST
make: Entering directory '/var/lib/dkms/chipsec/964.af7fa73/build'
make -C /lib/modules/`uname -r`/build M=/var/lib/dkms/chipsec/964.af7fa73/build clean
make[1]: Entering directory '/var/lib/dkms/chipsec/964.af7fa73/build'
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[1]: Leaving directory '/usr/lib/modules/5.7.7-arch1-1/build'
rm -f amd64/cpu.o
nasm -f elf64 -o amd64/cpu.o amd64/cpu.asm
make[1]: Entering directory '/var/lib/dkms/chipsec/964.af7fa73/build'
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
  CC [M]  /var/lib/dkms/chipsec/964.af7fa73/build/chipsec_km.o
[...]

PR #954, opened 2020-07-07
Fix for 'unexport of the kallsyms_lookup_name function' in kernels from 5.7.0 on

PaintedBlck:
Hi,

in follow up of topic _**Blackarch - DKMS make error #953**_, please find the patch I have written.

I have tried to be as less intrusive as possible. The functional logic of the program has not been changed. The only file that is patched is drivers/linux/chipsec_km.c

The patch is _**tested on Blackarch 5.7.7-arch1-1**_

_**!!! P L E A S E   T E S T   T H I S   P A T C H  !!!  on other Linux OS'es**_

Kind regards,
PaintedBlck

### DIFF
--- a/drivers/linux/chipsec_km.c
+++ b/drivers/linux/chipsec_km.c
@@ -30,6 +30,11 @@ chipsec@intel.com
 #include <linux/smp.h>
 #include <linux/miscdevice.h>
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,7,0)
+// Fix for 'unexport of the kallsyms_lookup_name function' in kernels from 5.7 on
+#include <linux/kprobes.h>
+#endif
+
 #include "include/chipsec.h"
 
 #ifdef HAS_EFI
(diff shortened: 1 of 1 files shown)


iherna2:
@PaintedBlck I tried this patch on a system I have running 5.7.0 and it fails to load the module. 

I added a couple of extra printk statements and the symbol that is failing to resolve is phys_mem_access_prot, register_kprobe returns error -22 (-EINVAL i believe).

I am guessing there is some config difference between the kernel you run and mine. Can you share your .config?

Do you have any other ideas to try?

PaintedBlck:
_**In reply of your question**_

I have put the kernel build `.config` in the github repo [PaintedBlck/chipsec-kernel_5_7_7-config](https://github.com/PaintedBlck/chipsec-kernel_5_7_7-config)

I have also joined the result of a `dkms status` to show that the fixed chipsec module, loads at startup of the kernel.

To make a final test easy, I have changed the /usr/src/chipsec-964.af7fa73/chipsec_km.c with my patched chipsec_km.c (which makes that version numbers of the module, etc. are not changed).

_**Possible problem**_

Would it be possible that the problem with the `kallsyms_lookup_name` is only present at version 5.7.7 and not at previous versions? Now, I'm interfering in all builds from version 5.7.0 on with three `#if` clauses. If so, my patch has to be adapted in the sence:

Replace all `#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,7,0)`
with `#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,7,7)`

iherna2:
@PaintedBlck, I've installed 5.7.8 and the module still fails loading. I need to look closely at the kernel config file you posted to find out what option(s) might be causing the symbols to not resolve.

PaintedBlck:
@iherna2 I have, unfortunately, encountered the same installation problem on the kernel 5.7.8

I will close this pull request and look further for a solution.

PR #957, opened 2020-07-08
A couple fixes to native mode

iherna2:
With this changes, I get mostly the same results on native vs with driver.

### DIFF
--- a/chipsec/helper/linux/linuxhelper.py
+++ b/chipsec/helper/linux/linuxhelper.py
@@ -83,10 +83,9 @@ class MemoryMapping(mmap.mmap):
     This subclass keeps tracks of the start and end of the mapping.
     """
     def __init__(self, fileno, length, flags, prot, offset):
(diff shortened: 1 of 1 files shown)


chipsec/helper/linux/linuxhelper.py
```diff
@@ -454,6 +453,7 @@ def native_write_pci_reg(self, bus, device, function, offset, value, size=4, dom
                 from chipsec.helper.linux.legacy_pci import LEGACY_PCI
                 pci = LEGACY_PCI()
                 value = pci.write_pci_config(bus, device, function, offset, value)
+                return False
```
npmitche, line 456: What was the reason behind returning False here? From what I can tell, most things just ignore a return value.

iherna2: Just to make it explict, I you feel against it, I can just drop the line

BrentHoltsclaw: My vote would be to leave it

PR #984, opened 2020-08-14
Additional help check in msr_cmd.py

frinzell:
- Additional check for less than 2 CLI elements.


npmitche:
Does this need to go into io_cmd and spidesc_cmd? https://github.com/chipsec/chipsec/commit/d7646897f798b5d6fe767833daa7474573487edd

frinzell:
Not needed.

PR #987, opened 2020-08-18
Only check cpuid for platform if driver is loaded

BrentHoltsclaw:


BrentHoltsclaw:
The alternative to this patch would be to add "native cpuid command" to OS other than Linux.  Any thoughts on which is better?

p-state:
@BrentHoltsclaw this might help https://github.com/flababah/cpuid.py

BrentHoltsclaw:
@p-state thanks. That code looks similar to what we are using for the native linux function.  It just needs to be ported/verified to work with other platforms.

PR #991, opened 2020-08-21
Added function to check if a register defines all fields in input list

npmitche:

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -679,6 +679,8 @@ class Chipset:
 #   reads/writes the value of the field (by name) of all configuration register instances (by register name)
 # register_has_field
 #   checks if the register has specific field
+# register_has_all_fields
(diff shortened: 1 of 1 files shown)


chipsec/chipset.py
```diff
@@ -933,6 +935,18 @@ def register_has_field( self, reg_name, field_name ):
             return False
         return (field_name in reg_def['FIELDS'])
 
+    def register_has_all_fields(self, reg_name, field_list):
+        try:
+            self.get_register_def(reg_name)
+        except KeyError:
+            return False
```
BrentHoltsclaw, line 942: it looks like lines 939-942 are repeated in register_has_field function.  Not sure if they are necessary.

npmitche: I think you are right. The way I was first wrote this, it would have returned True if the first iteration failed. I updated my loop and didn't think to take this out.

PR #1018, opened 2020-10-02
Scanner changes and updates

frinzell:


npmitche (approved):

0xfede7c8:
Are these only naming changes right?

frinzell:
Essentially.  Mostly comments, a few routine name changes, and pointer to updated default json.

0xfede7c8:
Ok, I will run some smoke tests to check that nothing breaks because of missing changes somewhere.

0xfede7c8:
> Ok, I will run some smoke tests to check that nothing breaks because of missing changes somewhere.

Run a few tests and did not detect anything new.

mjkyung (approved):

PR #1046, opened 2020-11-17
Fix chipsec.ko driver filepath for linux

nickylimjj:
In building Chipsec [rel 1.5.6](https://github.com/chipsec/chipsec/releases/tag/1.5.6) for Ubuntu 18.04 with `Linux 5.4.0-53-generic`, I had the following message when i ran `# python chipsec_main.py`

`ERROR: Message: "Cannot find chipsec.ko module"`

It seems that the file path was erroneous and i changed it in this PR.

### DIFF
--- a/chipsec/helper/linux/linuxhelper.py
+++ b/chipsec/helper/linux/linuxhelper.py
@@ -152,7 +152,7 @@ class LinuxHelper(Helper):
             else:
                 a2 = "a2=0x{}".format(phys_mem_access_prot)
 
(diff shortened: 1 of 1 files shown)


npmitche:
Hi! Thank you for the pull request! If you follow the wiki for [Installing on Linux](https://github.com/chipsec/chipsec/wiki/Installing-CHIPSEC-in-Linux) and run the command `python setup.py build_ext -i` this will copy the .ko file to the expected location.

nickylimjj:
Ah thanks for the clarification. I was only refering to the `manual.pdf` and didn't think to look at the wiki.

PR #1057, opened 2020-12-07
Add the support for access the kernel symbols by reading /proc/kallsy…

citypw:
Add the support for access the kernel symbols by reading /proc/kallsyms directly since kallsyms_lookup_name() and kallsyms_on_each_symbol() are no longer exported:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0bd476e6c67190b5eb7b6e105c8db8ff61103281

### DIFF
--- a/drivers/linux/chipsec_km.c
+++ b/drivers/linux/chipsec_km.c
@@ -66,6 +66,11 @@ module_param(a2,ulong,0); //a2 is addr of phys_mem_access_prot function
 /// Char we show before each debug print
 const char program_name[] = "chipsec";
 
+// read kernel symbols from the /proc
+#define KALLSYMS_PATH "/proc/kallsyms"
+#define BUFF_SIZE 128
+char read_buf[BUFF_SIZE] = {0};
+
 // list of allocated memory
 struct allocated_mem_list allocated_mem_list;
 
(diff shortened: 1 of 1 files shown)


npmitche (approved):

drivers/linux/chipsec_km.c
```diff
@@ -1676,6 +1681,94 @@ static struct miscdevice chipsec_dev = {
...
+				goto CLEAN_UP;
+			}
+
+			i = -1;
+			memset(proc_ksyms_entry, 0, BUFF_SIZE);
+		}
+
+		i++;
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,0,0)
```
mjkyung, line 1740: Is this condition ever happening, given the LINUX_VERSION_CODE here should be higher than Kerner version 5.4.0 already?

citypw: Those code should work on >= vanilla kernel v5.0.0 but it's clearly won't work under v5.4.0 since the commit is try to compatible w/ wider version of grsec. So what do you think if we change LINUX_VERSION_CODE to > v5.4.0( make sense to the current chipsec context) or just remain the original code there? I'm ok with both. Let me know your thought.

npmitche: I believe the issue is this "if" is inside the "if" on line 1688. If it's the kernel version is less than 5.4.0, it will use kallsyms_lookup_name, and skip this "if" all together.

citypw: Removed the conditional compilation due to its no use in chipsec. Please review.

BrentHoltsclaw (approved):

npmitche (approved):

drivers/linux/chipsec_km.c
npmitche: Awesome. Thank you! I merged it in and it will be included in the next release.

PR #1066, opened 2020-12-18
Expand the conditional compilation based on d02624b to support more P…

citypw:
Expand the conditional compilation based on d02624b to support more PaX/GRsec kernel

### DIFF
--- a/drivers/linux/chipsec_km.c
+++ b/drivers/linux/chipsec_km.c
@@ -560,11 +560,22 @@ int __weak phys_mem_access_prot_allowed(struct file *file,
         return 1;
 }
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,12) && defined(ARCH_HAS_VALID_PHYS_ADDR_RANGE) && !defined(__HAVE_ARCH_PAX_OPEN_USERLAND)
+#if defined(__HAVE_ARCH_PAX_OPEN_USERLAND)
+#if LINUX_VERSION_CODE > KERNEL_VERSION(4,14,0) && defined(ARCH_HAS_VALID_PHYS_ADDR_RANGE)
 int valid_mmap_phys_addr_range(unsigned long pfn, size_t size)
 {
 	return 1;
 }
+#endif
+
+#else
+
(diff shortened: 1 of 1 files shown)


drivers/linux/chipsec_km.c
```diff
@@ -560,13 +560,24 @@ int __weak phys_mem_access_prot_allowed(struct file *file,
...
+#if LINUX_VERSION_CODE > KERNEL_VERSION(4,14,0) && defined(ARCH_HAS_VALID_PHYS_ADDR_RANGE)
 int valid_mmap_phys_addr_range(unsigned long pfn, size_t size)
 {
 	return 1;
 }
 #endif
 
+#else
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,12) && defined(ARCH_HAS_VALID_PHYS_ADDR_RANGE)
```
mjkyung, line 573: Is there any specific reason why this condition is separate from from the one from L564?

drivers/linux/chipsec_km.c
```diff
@@ -560,13 +560,24 @@ int __weak phys_mem_access_prot_allowed(struct file *file,
         return 1;
 }
 
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,12) && defined(ARCH_HAS_VALID_PHYS_ADDR_RANGE) && !defined(__HAVE_ARCH_PAX_OPEN_USERLAND)
+#if defined(__HAVE_ARCH_PAX_OPEN_USERLAND)
```
mjkyung, line 563: From the previous version it is asking for !defined - is this a valid change?

citypw: it should be valid if you're using the grsec kernel > 4.14.

drivers/linux/chipsec_km.c
citypw: the idea was: If the kernel is PaX/Grsec > v4.14, it'll use the "return 1" hack since grsec seems didn't impl valid_mmap_phys_addr_range() by their own. Otherwise, it'll just keep what the CHIPSEC does.

citypw:
the code has been tested it on grsec-5.4.x and vanilla 5.9.x.

npmitche (approved):

frinzell (approved):

mjkyung (approved):

PR #1078, opened 2021-01-17
Fix dumping of CPU descriptor tables (GDT/IDT/LDT).

assafcarlsbad:


assafcarlsbad:
On Python 2 some tests are failing, while on Python 3 they complete successfully.
@BrentHoltsclaw is Python 2 still supported by chipsec?

npmitche:
As python 2 was end of lifed Jan 2020, we maintained support in the OSs till June 2020. From the release notes: 
For Windows, Linux and MacOS: Python2 support has been deprecated as of June 2020. Please use Python3.

At this time the only support we provide for python 2 is in the UEFI shell. 

That being said, can you share what tests you are seeing issues on?

assafcarlsbad:
See https://travis-ci.org/github/chipsec/chipsec/builds/759038474.

PR #1084, opened 2021-01-23
Update interrupts_cmd.py CLI

frinzell:
**ATTN CHIPSEC Community:**

**CALL FOR FEEDBACK**

In an effort to align the utill commands, **interrupts_cmd.py** is adding `send` command.  This will change the default behavior of a long established command-line invocation.

Any comments FOR or AGAINST, please post in comments.  

**For instance:**

Previous 'read' command looked like this...
`python chipsec_util.py smi 0x0 0xDE 0x0 `

Will now look like...
`python chipsec_util.py smi send 0x0 0xDE 0x0`

PR #1092, opened 2021-02-04
Convert interrupts_cmd to argparse

nstarke:
This commit attempts to convert the interrupts_cmd file
to use argparse (addresses #673).  Note that this commit
proposes a change to the smi command arguments by splitting
the default command into "sw" and "apmc" commands; if these
kinds of API changes are not possible this can still be
refactored to conform to the original API, albeit at the
slight cost of code cleanliness.

### DIFF
--- a/chipsec/utilcmd/interrupts_cmd.py
+++ b/chipsec/utilcmd/interrupts_cmd.py
@@ -24,6 +24,7 @@ import time
 
 from chipsec.command        import BaseCommand
 from chipsec.hal.interrupts import Interrupts
+from argparse               import ArgumentParser
 import os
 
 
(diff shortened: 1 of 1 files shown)


nstarke:
Ah I wasn't understanding the source code correctly - looks like this sort of API change will break a lot of stuff.  I also see that there is a proposal to change the API already in #1084 - I'll defer to that change.  Probably doesn't make sense to refactor this until #1084 is merged.

nstarke:
I didn't realize that most of this work to convert to argparse was done in #1084 already, I'm going to close this since its basically duplicated work.

npmitche:
If you have any feedback on how it was done in #1084, please leave a comment.

PR #1101, opened 2021-02-20
 Introduce a new module to check that SMM_Code_Chk_En is configured properly.

assafcarlsbad:
Typical output looks like this:

image

### DIFF
--- a/chipsec/cfg/8086/common.xml
+++ b/chipsec/cfg/8086/common.xml
@@ -616,6 +616,10 @@ Common (default) XML platform configuration file
       <field name="LOCK"           bit="30" size="1"  desc="Lock (set automatically on the first SMI)" />
       <field name="DEBUG_OCCURRED" bit="31" size="1"  desc="Debug Occurred (set by hardware)" />
     </register>
+    <register name="MSR_SMM_FEATURE_CONTROL" type="msr" msr="0x4E0" desc="Enhanced SMM Feature Control">
+      <field name="LOCK"            bit="0" size="1"  desc="Lock bit" />
+      <field name="SMM_Code_Chk_En" bit="2" size="1"  desc="Prevents SMM from executing code outside the ranges defined by the SMRR" />
+    </register>
 
     <!-- SGX related Model Specific Registers (MSR) -->
     <!-- Check if CPUID(EAX=7,ECX=0):EBX[2] is set before accessing these registers-->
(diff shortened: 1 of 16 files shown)


Sae86 (approved):

chipsec/cfg/8086/common.xml
```diff
@@ -616,6 +616,10 @@ Common (default) XML platform configuration file
       <field name="LOCK"           bit="30" size="1"  desc="Lock (set automatically on the first SMI)" />
       <field name="DEBUG_OCCURRED" bit="31" size="1"  desc="Debug Occurred (set by hardware)" />
     </register>
+    <register name="MSR_SMM_FEATURE_CONTROL" type="msr" msr="0x4E0" desc="Enhanced SMM Feature Control">
```
BrentHoltsclaw, line 619: I don't think this is actually a common msr, may want to update it in the configuration files which it applies.

chipsec/modules/common/smm_code_chk.py
```diff
@@ -0,0 +1,81 @@
+"""
```
BrentHoltsclaw, line 1: GPL header is needed

chipsec/modules/common/smm_code_chk.py
```diff
@@ -0,0 +1,81 @@
...
+        self.logger.start_test( "SMM_Code_Chk_En (SMM Call-Out) Protection" )
+
+        results = []
+        for tid in range(self.cs.msr.get_cpu_thread_count()):
+            results.append(self._check_SMM_Code_Chk_En(tid))
+
+        # Check that all CPUs have the same value of MSR_SMM_FEATURE_CONTROL.
+        if not all(_ == results[0] for _ in results):
+            self.logger.log_failed_check( "MSR_SMM_FEATURE_CONTROL does not have the same value across all CPUs" )
+            return ModuleResult.ERROR
```
BrentHoltsclaw, line 62:
```suggestion
            return ModuleResult.FAILED
```

chipsec/modules/common/smm_code_chk.py
```diff
@@ -0,0 +1,81 @@
...
+        
+        res = results[0] 
+        if res == ModuleResult.ERROR:
```
BrentHoltsclaw, line 65: Is this supposed to be FAILED opposed to ERROR?

chipsec/cfg/8086/common.xml
assafcarlsbad: Based on the Intel SDM it seems that the `MSR_SMM_FEATURE_CONTROL` register is available starting from:
* 4th Generation Intel® Core™ Processors (Haswell microarchitecture)
* Atom Processors Based on the Goldmont Microarchitecture

I moved the register definition to the appropriate files. Can you please double-check that I didn't make any mistakes?

BrentHoltsclaw (approved):

npmitche (approved):

PR #1102, opened 2021-02-21
Logger: prefer WConio2 over the original WConio.

assafcarlsbad:
`WConio2` is superior to the original `WConio` as it is more up-to-date and better maintained.
See https://pypi.org/project/WConio2/ for more details.

### DIFF
--- a/chipsec/logger.py
+++ b/chipsec/logger.py
@@ -34,11 +34,17 @@ import atexit
 from chipsec.testcase import TestCase, ChipsecResults
 import traceback
 try:
-    import WConio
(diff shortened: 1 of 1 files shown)


BrentHoltsclaw (approved):

BrentHoltsclaw:
Would it be better to drop Wconio support and just use Wconio2?

assafcarlsbad:
I thought it would be good to retain backward compatibility with the original `WConio`, but if you want to drop support for it then go ahead.

mjkyung (approved):

Sae86 (approved):

PR #1157, opened 2021-04-08
Add bytes support to uefivar_fuzz.py

frinzell:
- Support bytes name type
- Easier to invoke
- Minor formatting updates


ErikBjorge (approved):
The only comment here would be to do this in two commits on the same branch.  One to fix the issue and a second to refactor for code formatting.

mjkyung (approved):

Sae86 (approved):

PR #1163, opened 2021-04-18
Fix conversion of bytes buffers to strings

lucalandolfi:
This fixes a couple of exceptions that are raised when converting bytes buffers to strings. Tested with Python 3.9.2 on Debian 11.

* `chipsec/helper/linux/linuxhelper.py`
in `kern_get_EFI_variable_full()` changed `tostring()` to `tobytes()` because `tostring()` method of `array.array` has been deprecated since Python 3.9

* `chipsec/modules/tools/uefi/s3script_modify.py`
make sure the argument of `print_buffer()` is a string using `bytestostring()`

* `chipsec/hal/physmem.py`
make sure the argument of `print_buffer()` is a string using `bytestostring()`

* `chipsec/modules/tools/vmm/common.py`
in `get_initial_data()` removed `ord()` from strings construction because `DD()` returns a `bytes` object

* `chipsec/modules/common/uefi/access_uefispec.py`
in `diff_var()` removed `ord()` call because `data1` and `data2` variables are `bytes` objects

PR #1170, opened 2021-04-20
Change tostring to tobytes in linuxhelper.py

dnoliver:
Tested with Python3.7 and Python3.9

Python docs at https://docs.python.org/3/whatsnew/3.9.html#removed says that `tostring` was an alias to `tobytes` anyways.


ErikBjorge (approved):
Same change included in #1163.

BrentHoltsclaw (approved):

frinzell (approved):

mjkyung (approved):

PR #1134, opened 2021-03-25
Adding read_all_register_instances() method

iherna2:
This method returns a list containig the values of all the
instances found of the requested register.

In the case of MSRs it reads the scope of the MSR from platform
configuration xml.

Signed-off-by: Ignacio Hernandez <ignacio.hernandez@intel.com>

PR #1175, opened 2021-05-01
Extend register support to cs.read/write_all functions

BrentHoltsclaw:
included changes from #1134


BrentHoltsclaw:
Have a better fix, closing this PR

PR #1182, opened 2021-05-10
Fix issue when trying to access SB on some platforms

npmitche:


chipsec/hal/msgbus.py
```diff
@@ -89,6 +89,7 @@ class MsgBus(hal_base.HALBase):
     def __init__(self, cs):
         super(MsgBus, self).__init__(cs)
         self.helper = cs.helper
+        self.p2sbHide = None
```
frinzell, line 92: Empty directory might be better.

frinzell (approved):

Sae86 (commented):
Looks good

PR #1190, opened 2021-05-18
Fix unassigned variable

BrentHoltsclaw:


npmitche (approved):

Sae86 (approved):

iherna2 (changes requested):
line 193 should be changed to use 'preserve' variable as well

npmitche (approved):

PR #1201, opened 2021-05-27
Enhance multibus support

BrentHoltsclaw:
Add mulibus support to mmio.hal
Update chipset.cs functions to add multibus mmio support
Update chipset.cs to allow read/write all functions to support msr and mmio

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -903,7 +903,8 @@ class Chipset:
         elif RegisterType.MMCFG == rtype:
             self.mmio.write_mmcfg_reg(int(reg['bus'], 16), int(reg['dev'], 16), int(reg['fun'], 16), int(reg['offset'], 16), int(reg['size'], 16), reg_value )
         elif RegisterType.MMIO == rtype:
-            self.mmio.write_MMIO_BAR_reg(reg['bar'], int(reg['offset'], 16), reg_value, int(reg['size'], 16), bus)        elif RegisterType.MSR == rtype:
+            self.mmio.write_MMIO_BAR_reg(reg['bar'], int(reg['offset'], 16), reg_value, int(reg['size'], 16), bus)
+        elif RegisterType.MSR == rtype:
             eax = (reg_value & 0xFFFFFFFF)
             edx = ((reg_value >> 32) & 0xFFFFFFFF)
             self.msr.write_msr( cpu_thread, int(reg['msr'], 16), eax, edx )
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -699,9 +699,9 @@ class Chipset:
                 d = int(reg['dev'], 16)
                 f = int(reg['fun'], 16)
                 return self.pci.is_enabled( b, d, f )
-            if (rtype == RegisterType.MMIO):
+            elif (rtype == RegisterType.MMIO):
                 bar_name = reg['bar']
-                return self.mmio.is_MMIO_BAR_enabled(bar_name)
+                return self.mmio.is_MMIO_BAR_enabled(bar_name, bus)
         return False
 
     def switch_device_def( self, target_dev, source_dev ):
@@ -832,7 +832,7 @@ class Chipset:
             reg_value = self.mmio.read_mmcfg_reg(int(reg['bus'], 16), int(reg['dev'], 16), int(reg['fun'], 16), int(reg['offset'], 16), int(reg['size'], 16) )
         elif RegisterType.MMIO == rtype:
             if self.mmio.get_MMIO_BAR_base_address(reg['bar'])[0] != 0:
-                reg_value = self.mmio.read_MMIO_BAR_reg(reg['bar'], int(reg['offset'], 16), int(reg['size'], 16) )
+                reg_value = self.mmio.read_MMIO_BAR_reg(reg['bar'], int(reg['offset'], 16), int(reg['size'], 16), bus)
             else:
                 raise CSReadError("MMIO Bar ({}) base address is 0".format(reg['bar']))
         elif RegisterType.MSR == rtype:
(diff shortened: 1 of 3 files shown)


BrentHoltsclaw:
Depends upon #1200

chipsec/chipset.py
```diff
@@ -912,24 +927,59 @@ def write_register(self, reg_name, reg_value, cpu_thread=0, bus_index=0):
...
+                threads_to_use = range(self.helper.get_threads_count())
+            if len(reg_values) == len(threads_to_use):
+                value = 0
+                for t in threads_to_use:
+                    self.write_register(reg_name, reg_values[value], t)
+                    value += 1
+                ret = True
+        elif rtype in [RegisterType.MMCFG, RegisterType.PCICFG, RegisterType.MMIO] and bus_data:
+            values = len(bus_data)
+            if len(reg_values) == values:
```
iherna2, line 952: This can become problematic for generalizing tests applicable to single and multi-socket, it means that the user would need to find out what is the current system topology before calling this method, making it less useful. Would it help to make it possible to program a single value to every instance?

chipsec/chipset.py
```diff
@@ -912,24 +927,59 @@ def write_register(self, reg_name, reg_value, cpu_thread=0, bus_index=0):
...
+                threads_to_use = [cores[p][0] for p in cores]
+            else: # Default to threads
+                threads_to_use = range(self.helper.get_threads_count())
+            for t in threads_to_use:
+                self.write_register(reg_name, reg_value, t)
+        elif rtype in [RegisterType.MMCFG, RegisterType.PCICFG, RegisterType.MMIO] and bus_data:
+            for bus in bus_data:
+                self.write_register(reg_name, reg_value, cpu_thread, bus)
+        else:
+            self.write_register(reg_name, reg_value)
```
iherna2, line 982: Back to the previous note, if we have the explicit method to write a single value to all instances of a register, I am questioning even more the usefulness of the write_register_all()

chipsec/chipset.py
BrentHoltsclaw: The expectation would be that the read_register_all is called, and then modify the fields before using the write_register_all command.

iherna2 (approved):

chipsec/chipset.py
```diff
@@ -912,24 +928,59 @@ def write_register(self, reg_name, reg_value, cpu_thread=0, bus_index=0):
...
+        return ret
 
     def write_register_all_single(self, reg_name, reg_value, cpu_thread=0):
-        bus_data = self.get_register_bus( reg_name )
-        if not bus_data:
-            return False
-        values = len(bus_data)
-        for index in range( values ):
-            self.write_register( reg_name, reg_value, cpu_thread, index )
+        reg = self.get_register_def(reg_name)
```
frinzell, line 964: Would it work if this pointed to `write_register_all()` for when `len(reg_values)` == 1.

chipsec/chipset.py
frinzell: How about a verbose (HAL?) log if `len(reg_values)` > 1 and != `len(bus_data)`?

chipsec/hal/mmio.py
```diff
@@ -190,10 +190,13 @@ def get_MMIO_BAR_base_address(self, bar_name):
                     reg_mask = 0xFFFF
             else:
                 base = self.cs.read_register(bar_reg)
-                reg_mask = self.cs.get_register_field_mask(bar_reg, preserve)
+                reg_mask = self.cs.get_register_field_mask(bar_reg, preserve_field_position=preserve)
         else:
             # this method is not preferred (less flexible)
-            b = int(bar['bus'], 16)
+            if bus:
```
frinzell, line 196: `bus` = 0 could be valid, might need to check for `is not None`.

chipsec/hal/mmio.py
```diff
@@ -310,26 +318,33 @@ def dump_MMIO_BAR(self, bar_name):
...
+
+                if 'register' in _bar:
+                    _s = _bar['register']
+                    if 'offset' in _bar:
+                        _s += (' + 0x{:X}'.format(int(_bar['offset'], 16)))
+                else:
+                    _s = '{:02X}:{:02X}.{:01X} + {}'.format( int(_bar['bus'], 16), int(_bar['dev'], 16), int(_bar['fun'], 16), _bar['reg'] )
 
-            self.logger.log( ' {:12} | {:14} | {:016X} | {:08X} | {:d}   | {}'.format(_bar_name, _s, _base, _size, _en, _bar['desc']) )
+                self.logger.log( ' {:12} | {:02X} | {:14} | {:016X} | {:08X} | {:d}   | {}'.format(_bar_name, bus, _s, _base, _size, _en, _bar['desc']) )
```
frinzell, line 347: Need one more space in the BUS column to align with header.

chipsec/hal/mmio.py
```diff
@@ -310,26 +318,33 @@ def dump_MMIO_BAR(self, bar_name):
...
-
+            bus_data = None
             if 'register' in _bar:
-                _s = _bar['register']
-                if 'offset' in _bar: _s += (' + 0x{:X}'.format(int(_bar['offset'], 16)))
-            else:
-                _s = '{:02X}:{:02X}.{:01X} + {}'.format( int(_bar['bus'], 16), int(_bar['dev'], 16), int(_bar['fun'], 16), _bar['reg'] )
+                bus_data = self.cs.get_register_bus(_bar['register'])
+            if bus_data is None:
+                bus_data = [0]
```
frinzell, line 331: Is it better to get bus through `get_register_def()`?

chipsec/chipset.py
BrentHoltsclaw: can you clarify this comment?

chipsec/chipset.py
BrentHoltsclaw: It does not look like that would work

chipsec/chipset.py
frinzell: If `reg_values` doesn't align with `bus_data`, let the user know.  Might help with @iherna2 point about non-aligning topology.

BrentHoltsclaw: Do you want an exception or log?

BrentHoltsclaw: Because the function returns false if the values are not written a log seems appropriate

npmitche: Is this where you were going to add the exception.

frinzell (approved):

npmitche (approved):

PR #1218, opened 2021-06-22
Warn and force user to interact if running python2 in an OS

npmitche:


chipsec/helper/oshelper.py
```diff
@@ -74,6 +74,13 @@ def __init__(self):
             os_system  = platform.system()
             raise OsHelperError( "Could not load any helpers for '{}' environment (unsupported environment?)".format(os_system), errno.ENODEV )
         else:
+            if self.helper.name != "EfiHelper" and sys.version[0] == "2": # Python 2 is only supported in the EFI shell.
+                logger().warn("***************************************************************************************")
+                logger().warn("* !! Python 2 is deprecated. Please update to Python 3 !!")
+                logger().warn("* Some chipsec results may be incorrect if you continue.")
+                logger().warn("***************************************************************************************")
+                s = raw_input( "Type 'yes' to continue running under Python 2 > " ) # Will only run on python 2, so raw_input will be defined.
+                if s != 'yes': sys.exit( 0 )
```
frinzell, line 83: Is 'y' acceptable?

fishilico:
For information, the continuous integration checks currently fail because they test Python 2.7 on Ubuntu (cf. https://github.com/chipsec/chipsec/blob/1.6.4/.github/workflows/tests.yml#L177-L183). The lines with `python: 2.7` need to be removed from `.github/workflows/tests.yml` in order to no longer test Python 2.7 in Github Actions CI.

mjkyung (approved):

npmitche:
> For information, the continuous integration checks currently fail because they test Python 2.7 on Ubuntu (cf. https://github.com/chipsec/chipsec/blob/1.6.4/.github/workflows/tests.yml#L177-L183). The lines with `python: 2.7` need to be removed from `.github/workflows/tests.yml` in order to no longer test Python 2.7 in Github Actions CI.

There was another issue with the ubuntu 18 and 16 tests. When you run sudo python, it drops out of the python github action environment and defaulted back to python2. Took me a few tries to get it working.

frinzell (approved):

minipli-oss:
As already noticed by @fishilico above and @npmitche in https://github.com/chipsec/chipsec/pull/1228#issuecomment-889225154, this commit has great regression potential as in breaking existing scripted environments.

Instead of waiting for the user to confirm -- which a script won't, less so with its output redirected -- how about just emitting a deprecation warning? And, after a few releases, change the shebang to `python3` explicitly?

npmitche:
We have been messaging that python2 is deprecated for the past year in our release notes and most users have upgraded. Regarding the shebang, that would only affect linux users, and only those executing the script without explicitly calling a python binary. This PR is a stop gap to remove a debug step from our end.

minipli-oss:
> We have been messaging that python2 is deprecated for the past year in our release notes and most users have upgraded.

Fair enough. However, not all Linux distros are that fast with switching the python interpreter behind the version agnostic `python` symlink. Case in point, Debian buster, the current stable release, still makes it point to `python2` which is itself a symlink to `python2.7`.

> Regarding the shebang, that would only affect linux users, and only those executing the script without explicitly calling a python binary.

As I am, in fact, a Linux user and am executing chipsec's scripts by directly invoking `chipsec_main.py` / `chipsec_util.py` I am affected by that change and that's the reason why I do care about this PR. Maybe I'm the only chipsec user running it under Linux ;) In this case, changing the shebang wouldn't make a difference for you and others but would fix my setup. :)

> This PR is a stop gap to remove a debug step from our end.

What exactly do you mean by this? What kind of debug step?

minipli-oss:
Oh, and if you switch to python3 for real, you should update `debian/control` as well. Right now it lists a version dependency to python >= 2.6.

npmitche:
> Fair enough. However, not all Linux distros are that fast with switching the python interpreter behind the version agnostic `python` symlink. Case in point, Debian buster, the current stable release, still makes it point to `python2` which is itself a symlink to `python2.7`.

In your experience with other distros, when python3 is installed is there always a symlink/binary behind /bin/usr/python3? 

> As I am, in fact, a Linux user and am executing chipsec's scripts by directly invoking `chipsec_main.py` / `chipsec_util.py` I am affected by that change and that's the reason why I do care about this PR. Maybe I'm the only chipsec user running it under Linux ;) In this case, changing the shebang wouldn't make a difference for you and others but would fix my setup. :)

This is good feedback, thank you. The majority of customers we interact with run under Windows, and those running Linux generally explicitly call the python binary. I'll look into updating the #!.

> What exactly do you mean by this? What kind of debug step?

We often get emails saying "This is broken, what's wrong?" and one of the first steps we take is looking at the platform/environment it's being run on and python2 will on occasion come up. This will save us a step of "Please run using python3."

> Oh, and if you switch to python3 for real, you should update debian/control as well. Right now it lists a version dependency to python >= 2.6.

Thank you for pointing this out!

minipli-oss:
> > Fair enough. However, not all Linux distros are that fast with switching the python interpreter behind the version agnostic `python` symlink. Case in point, Debian buster, the current stable release, still makes it point to `python2` which is itself a symlink to `python2.7`.
> 
> In your experience with other distros, when python3 is installed is there always a symlink/binary behind /usr/bin/python3?

I did a few tests here:
- Arch Linux:
   ```sh-session
   [root@arch ~]# which python{,3}
   /usr/bin/python
   /usr/bin/python3
   [root@arch ~]# python --version
   Python 3.8.1
   ```
- Debian buster:

- SuSE Tumbleweed:
   ```sh-session
   tumbleweed:~ # which python{,3}
   /usr/bin/python
   /usr/bin/python3
   tumbleweed:~ # python --version
   Python 2.7.18
   ```

- Ubuntu 16.04:
   ```sh-session
   ```

- Ubuntu 16.04:
   ```sh-session
   ```
- 

> 
> > As I am, in fact, a Linux user and am executing chipsec's scripts by directly invoking `chipsec_main.py` / `chipsec_util.py` I am affected by that change and that's the reason why I do care about this PR. Maybe I'm the only chipsec user running it under Linux ;) In this case, changing the shebang wouldn't make a difference for you and others but would fix my setup. :)
> 
> This is good feedback, thank you. The majority of customers we interact with run under Windows, and those running Linux generally explicitly call the python binary. I'll look into updating the #!.
> 
> > What exactly do you mean by this? What kind of debug step?
> 
> We often get emails saying "This is broken, what's wrong?" and one of the first steps we take is looking at the platform/environment it's being run on and python2 will on occasion come up. This will save us a step of "Please run using python3."
> 
> > Oh, and if you switch to python3 for real, you should update debian/control as well. Right now it lists a version dependency to python >= 2.6.
> 
> Thank you for pointing this out!

minipli-oss:
> > Fair enough. However, not all Linux distros are that fast with switching the python interpreter behind the version agnostic `python` symlink. Case in point, Debian buster, the current stable release, still makes it point to `python2` which is itself a symlink to `python2.7`.
> 
> In your experience with other distros, when python3 is installed is there always a symlink/binary behind /usr/bin/python3?

I did a few tests here:

- Arch Linux:
   ```sh-session
   [root@arch ~]# which python{,3}
   /usr/bin/python
   /usr/bin/python3
   [root@arch ~]# python --version
   Python 3.8.1
   ```

- Alpine Linux 3.10:
   ```sh-session
   alpine:~# which python{,3}
   /usr/bin/python3
   alpine:~# python3 --version
   Python 3.7.10
   ```
- Amazon Linux 2 AMI:

- Debian buster:
   ```sh-session
   minipli@bell:~$ which python{,3}
   /usr/bin/python
   /usr/bin/python3
   minipli@bell:~$ python --version
   Python 2.7.16
   ```
- SuSE Tumbleweed:
   ```sh-session
   tumbleweed:~ # which python{,3}
   /usr/bin/python
   /usr/bin/python3
   tumbleweed:~ # python --version
   Python 2.7.18
   ```
- Ubuntu 16.04:
   ```sh-session
   root@ubuntu-1604:~# which python{,3}
   /usr/bin/python
   /usr/bin/python3
   root@ubuntu-1604:~# python --version
   Python 2.7.12
   ```
- Ubuntu 18.04:
   ```sh-session
   root@ubuntu-1804:~# which python{,3}
   /usr/bin/python3
   root@ubuntu-1804:~# python3 --version
   Python 3.6.9
   ```

So every distro beside AMI has a Python 3 interpreter available as `python3`. A few ship Python 3 only by default (Alpine, Ubuntu 18.04). Of those who ship both, the version behind the `python` shortcut is mixed (x: Python 2, y: Python 3).

> 
> > As I am, in fact, a Linux user and am executing chipsec's scripts by directly invoking `chipsec_main.py` / `chipsec_util.py` I am affected by that change and that's the reason why I do care about this PR. Maybe I'm the only chipsec user running it under Linux ;) In this case, changing the shebang wouldn't make a difference for you and others but would fix my setup. :)
> 
> This is good feedback, thank you. The majority of customers we interact with run under Windows, and those running Linux generally explicitly call the python binary. I'll look into updating the #!.
> 
> > What exactly do you mean by this? What kind of debug step?
> 
> We often get emails saying "This is broken, what's wrong?" and one of the first steps we take is looking at the platform/environment it's being run on and python2 will on occasion come up. This will save us a step of "Please run using python3."
> 
> > Oh, and if you switch to python3 for real, you should update debian/control as well. Right now it lists a version dependency to python >= 2.6.
> 
> Thank you for pointing this out!

minipli-oss:
> > Fair enough. However, not all Linux distros are that fast with switching the python interpreter behind the version agnostic `python` symlink. Case in point, Debian buster, the current stable release, still makes it point to `python2` which is itself a symlink to `python2.7`.
> 
> In your experience with other distros, when python3 is installed is there always a symlink/binary behind /usr/bin/python3?

I did a few tests here:

- Arch Linux:
   ```sh-session
   [root@arch ~]# which python{,3}
   /usr/bin/python
   /usr/bin/python3
   [root@arch ~]# python --version
   Python 3.8.1
   ```

- Alpine Linux 3.10:
   ```sh-session
   alpine:~# which python{,3}
   /usr/bin/python3
   alpine:~# python3 --version
   Python 3.7.10
   ```
- Amazon Linux 2 AMI:
   ```sh-session
   [root@ami ~]# which python{,3}
   /usr/bin/python
   /usr/bin/which: no python3 in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/root/bin)
   [root@ami ~]# python --version
   Python 2.7.18
   ```
- Debian buster:
   ```sh-session
   minipli@bell:~$ which python{,3}
   /usr/bin/python
   /usr/bin/python3
   minipli@bell:~$ python --version
   Python 2.7.16
   ```
- SuSE Tumbleweed:
   ```sh-session
   tumbleweed:~ # which python{,3}
   /usr/bin/python
   /usr/bin/python3
   tumbleweed:~ # python --version
   Python 2.7.18
   ```
- Ubuntu 16.04:
   ```sh-session
   root@ubuntu-1604:~# which python{,3}
   /usr/bin/python
   /usr/bin/python3
   root@ubuntu-1604:~# python --version
   Python 2.7.12
   ```
- Ubuntu 18.04:
   ```sh-session
   root@ubuntu-1804:~# which python{,3}
   /usr/bin/python3
   root@ubuntu-1804:~# python3 --version
   Python 3.6.9
   ```

So every distro beside AMI has a Python 3 interpreter available as `python3`. It probably simply isn't installed by default on AMI.
A few of the distros  ship Python 3 only by default (Alpine, Ubuntu 18.04).
Of those who ship the generic `python` shortcut, the used version behind it is mixed: 4 point to Python 2, 1 points to Python 3.

> > As I am, in fact, a Linux user and am executing chipsec's scripts by directly invoking `chipsec_main.py` / `chipsec_util.py` I am affected by that change and that's the reason why I do care about this PR. Maybe I'm the only chipsec user running it under Linux ;) In this case, changing the shebang wouldn't make a difference for you and others but would fix my setup. :)
> 
> This is good feedback, thank you. The majority of customers we interact with run under Windows, and those running Linux generally explicitly call the python binary. I'll look into updating the #!.

Thanks!

> > What exactly do you mean by this? What kind of debug step?
> 
> We often get emails saying "This is broken, what's wrong?" and one of the first steps we take is looking at the platform/environment it's being run on and python2 will on occasion come up. This will save us a step of "Please run using python3."

I see. Yeah, switching the #! will solve this probably for good.

> > Oh, and if you switch to python3 for real, you should update debian/control as well. Right now it lists a version dependency to python >= 2.6.
> 
> Thank you for pointing this out!

Sure, you're welcome!

npmitche:
We'll hold off on this PR WRT our next release and discuss in my team about having it accompanied with a #! update in the near future.  

Thanks again for your help!

fishilico:
> Regarding the shebang, that would only affect linux users, and only those executing the script without explicitly calling a python binary.

For information, I am also such a user, using for example `chipsec_util.py iommu pt` to dump the IOMMU tables.

> In your experience with other distros, when python3 is installed is there always a symlink/binary behind `/usr/bin/python3`?

This is also documented by Python maintainers in PEP 394 (https://www.python.org/dev/peps/pep-0394/). By the way, for scripts which may be run in Python virtual environments (created for example with `python3 -m venv my_venv ; . ./my_venv/bin/activate`), it is recommended to resolve `python3` in `$PATH`, using `/usr/bin/env`. More precisely instead of changing to `#!/usr/bin/python3`, I would suggest to change to:

```python
#!/usr/bin/env python3
```

(and this is already what is used in `chipsec_main.py` and `chipsec_util.py`. My point is to indicate that keeping `/usr/bin/env` is useful)

kerneis-anssi:
> For information, I am also such a user, using for example chipsec_util.py iommu pt to dump the IOMMU tables.

And for the record, I stumbled upon #1236 precisely because I also relied on the shebang to find the correct Python interpreter. Glad to see this fixed in upcoming releases.

PR #1228, opened 2021-07-09
Add log_error() routine

frinzell:


minipli-oss:
What's the purpose of adding yet another log helper function without a user? IMHO there are *way to many* log_*() helpers already. We should rather try to consolidate the existing ones instead of introducing new (unused) ones.

frinzell:
1. Log routine naming consistency
2. That's the goal

minipli-oss:
Lets provide some statistics to get an overview!

Of the 28 logging related methods of the `logger.py` module the following grep-based search gives us a rough number of their usage:

```sh-session
$ git grep -Eoh '\.(log|error|warn|verbose_log|log_passed_check|log_failed_check|log_error_check|log_skipped_check|log_warn_check|log_information_check|log_not_applicable_check|log_passed|log_failed|log_warning|log_skipped|log_not_applicable|log_heading|log_important|log_result|log_bad|log_good|log_unknown|log_information|start_test|start_module|end_module|_write_log|_save_to_log_file) *\(' -- "*.py" | tr -d '.(' | sort | uniq -c | sort -nr
   1654 log
    263 error
    107 warn
     91 log_bad
     70 log_good
     52 start_test
     32 log_passed_check
     28 log_important
     23 log_warn_check
     23 log_failed_check
     23 log_failed
     16 log_warning
      9 log_skipped_check
      9 log_passed
      5 log_information
      3 log_skipped
      3 log_information_check
      2 log_unknown
      2 log_not_applicable
      2 log 
      1 _write_log
      1 verbose_log
      1 start_module
      1 log_result
      1 log_not_applicable_check
      1 log_error_check
      1 end_module
```

Top user is by far the generic `log()` method, followed by `error()` and `warn()` which seems also rather intuitive. Similar in function, but longer in typing methods, e.g. `log_warn_check()` which just wraps `log_warning()` which differs from `warn()` only in cosmetic aspects, have a far lower usage count.

So IMHO these statistics clearly show that the ambiguity about the various similar named, but partially differently implemented logging methods actually *prevents* people from using the "correct" one and makes them stick to the generic `log()` method instead. There's clearly a need for a more compact and intuitive to use API. Having multiple methods for the same thing just causes confusion. We really should *remove* methods instead of adding new ones, to provide such a clean and intuitive API.

For example, the `*_check()` methods are just wrappers around their non-`_check` counterparts. Their intended usage seems to be to give a final test status message. However, neither is the "final" part enforced anywhere nor is there any special handling for the `*_check` variants. This gives us the opportunity to just drop the `*_check` ones as they're just bloat without additional functionality.

Site note observations:

1. The single caller of `_write_log()` is `_save_to_log_file()` -- which has no caller itself, vanishing the need for these two methods to exist at all.

2. The single user of `log_not_applicable_check()` is the SGX module, trying to differentiate `ModuleResult.SKIPPED` and `ModuleResult.NOTAPPLICABLE`:
   ```sh-session
   $ git grep -C1 'test is being skipped'
   chipsec/modules/common/sgx_check.py-            elif self.res == ModuleResult.SKIPPED:
   chipsec/modules/common/sgx_check.py:                self.logger.log_skipped_check('SGX test is being skipped')
   chipsec/modules/common/sgx_check.py-            elif self.res == ModuleResult.NOTAPPLICABLE:
   chipsec/modules/common/sgx_check.py:                self.logger.log_not_applicable_check('SGX test is being skipped')
   chipsec/modules/common/sgx_check.py-            else:
   ```
   However, the communicated log messages is the same for both cases -- to skip the test. So these could probably be combined, vanishing the need for yet another logging method.

3. The double mentioning of `log` in the above stats -- once with 1654 call sites, once with only 2 -- is actually a white space error for the latter case:
    ```sh-session
   $ git grep '\.log ('
   chipsec/hal/spi.py:                self.logger.log ( "  * Didn't find a valid SFDP header for Flash device {:d}".format(component +1))
   chipsec_main.py:            logger().log ('[CHIPSEC] *****************************************************************')
   ```

Sae86 (approved):

mjkyung (approved):

npmitche:
@frinzell Did we want to change the character in the brackets for the error? 

@minipli-oss The plan is to remove all of the *_check functions, we are just moving slowly.

minipli-oss:
But given the above usage distribution, we probably only need three logging functions at most: log(), warn() and error(). That should be sufficient to cover all use cases, as log() is the top one user anyway. All the rest can be mapped to one of these three, removing the need for more specialized logger functions with too vague and overlapping documentation about their intended usage.
Heck, we could even do with log() alone and do the markup based on the log level. When there's just one logging function it's hard to use wrong.

npmitche:
We have a lot of users that use chipsec in an automated environment, if we just had log() and left it up to the implementation to do markup, consistency would suffer and make harder or outright break automation.

chipsec/logger.py
```diff
@@ -280,6 +280,11 @@ def log_failed( self, text ):
         text = "[-] FAILED: " + text
         self.log(text, pyLogging.ERROR)
 
+    def log_error( self, text ):
+        """Logs an Error message"""
+        text = "[!] ERROR: " + text
```
npmitche, line 285:
```suggestion
        text = "[-] ERROR: " + text
```

npmitche (changes requested):

npmitche (approved):

minipli-oss:
> We have a lot of users that use chipsec in an automated environment, if we just had log() and left it up to the implementation to do markup, consistency would suffer and make harder or outright break automation.

What I meant was to add the "[-] ERROR: " / "[!] WARNING: " /... prefix based on the log level passed to the log() function. This would still support the script based processing of log messages. Even better, I'd say, as there would be only one type of error / warning / info messages, not multiple ones, each with different prefixes as it is right now.

PR #1235, opened 2021-08-02
Create exceptions.py for chipsec exceptions

BrentHoltsclaw:
Create a new exceptions.py file at the chipsec level to hold all chipsec exceptions.
This will help to avoid circular dependencies in the future.

Signed-off-by: brentholtsclaw <brent.holtsclaw@intel.com>

PR #1239, opened 2021-08-05
Create hal.locks and util_cmd.lock_check_cmd

BrentHoltsclaw:
Create
hal.locks - functionality to gather state information based upon locks defined within a configuration
util_cmd.lock_check_cmd - Use the hal locks functionality to print out the state of one or more locks identified within the configuration

Modify
chipset.py - fix up get_lock and set_lock functionality
hal.mmio.py - raise CSReadError exception when cannot read the MMIOBAR opposed to generic exception
hal.iobar - raise CSReadError exception when cannot read the IOBar
defines - add bit and is_set functions

Depends on #1235 

### DIFF
--- a/chipsec/hal/locks.py
+++ b/chipsec/hal/locks.py
@@ -26,11 +26,11 @@ from chipsec.exceptions import CSReadError, HWAccessViolationError
 
 
 class LockResult:
-    DEFINED = bit(0)
-    HAS_CONFIG = bit(1)
-    LOCKED = bit(2)
-    CAN_READ = bit(3)
-    INCONSISTENT = bit(4)
+    DEFINED = bit(0)  # lock exists within configuration
+    HAS_CONFIG = bit(1)  # lock configuration exists
+    LOCKED = bit(2)  # lock matches value within xml
+    CAN_READ = bit(3)  # system is able to access the lock
+    INCONSISTENT = bit(4)  # all lock results do not match
 
 
 class locks(HALBase):
@@ -49,15 +49,15 @@ class locks(HALBase):
             res |= LockResult.DEFINED
         try:
             self.cs.get_locked_value(lock_name)
-            self.cs.get_lock(lock_name, bus=bus)
             res |= LockResult.HAS_CONFIG
+            self.cs.get_lock(lock_name, bus=bus)
             res |= LockResult.CAN_READ
         except KeyError:
             pass
         except CSReadError:
-            res |= LockResult.HAS_CONFIG
+            pass
         except HWAccessViolationError:
-            res |= LockResult.HAS_CONFIG
+            pass
         return res
 
     def is_locked(self, lock_name, bus=None):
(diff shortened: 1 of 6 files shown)


chipsec/hal/locks.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def lock_valid(self, lock_name, bus=None):
+        res = 0
+        if self.cs.is_lock_defined(lock_name):
+            res |= LockResult.DEFINED
+        try:
+            self.cs.get_locked_value(lock_name)
+            self.cs.get_lock(lock_name, bus=bus)
+            res |= LockResult.HAS_CONFIG
+            res |= LockResult.CAN_READ
```
npmitche, line 54: Should lines 52 and 53 be swapped? (Set `HAS_CONFIG` before trying to get the lock) Then for `CSReadError` and `HWAccessViolationError`, you can just `pass` instead of setting it in the exception handler?

chipsec/hal/locks.py
```diff
@@ -0,0 +1,75 @@
...
+        return res
+
+    def is_locked(self, lock_name, bus=None):
+        """
+        Return whether the lock has the value setting
+        """
+        res = self.lock_valid(lock_name, bus)
+        if is_set(res, LockResult.HAS_CONFIG) and is_set(res, LockResult.CAN_READ):
+            locked = self.cs.get_locked_value(lock_name)
+            lock_setting = self.cs.get_lock(lock_name, bus=bus)
```
npmitche, line 70: Is there a reason why you're calling these functions again and not simply passing back the values in the `lock_valid` function?

chipsec/utilcmd/lock_check_cmd.py
```diff
@@ -0,0 +1,159 @@
...
+            res = "{}\n{}".format(res, self.check_log(lock, is_locked))
+        if self.logger.HAL:
+            self.logger.log(res)
+        return
+
+    def check_lock(self):
+        res = self.log_header()
+        if self.logger.VERBOSE:
+            self.log_key()
+        for lock in self.lockname:
```
npmitche, line 113: You don't have it documented anywhere that you allow multiple locks to be inputted in a single command line.

chipsec/hal/locks.py
BrentHoltsclaw: `lock_valid` checks to see if the configuration is correct and the register can be accessed.  Sets the appropriate LockResult flags for DEFINED (lock exists within configuration), HAS_CONFIG (lock has full configuration), and CAN_READ (lock can be accessed by the system).
`is_locked`  compares the value within xml to the register value. Sets the appropriate LockResult flags for LOCKED (lock value matches system read value) and INCONSISTENT (in cases of multibus or msr the register is not set the same)

npmitche: Makes sense logically, but my optimization brain is arguing it's wasting cycles.

Sae86 (approved):

PR #1302, opened 2021-11-11
Added ability to dump unnamed MMIO regions

ErikBjorge:

### DIFF
--- a/chipsec/utilcmd/mmio_cmd.py
+++ b/chipsec/utilcmd/mmio_cmd.py
@@ -21,15 +21,21 @@
 """
 >>> chipsec_util mmio list
 >>> chipsec_util mmio dump <MMIO_BAR_name> [offset] [length]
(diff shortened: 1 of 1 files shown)


BrentHoltsclaw (approved):

chipsec/utilcmd/mmio_cmd.py
```diff
@@ -21,15 +21,21 @@
 """
 >>> chipsec_util mmio list
 >>> chipsec_util mmio dump <MMIO_BAR_name> [offset] [length]
+>>> chipsec_util mmio dump-abs <MMIO_base_address> <length>
```
npmitche, line 24: Can/should we include an offset in the dump-abs command? Seems like all the abs should match, either all have it or all don't?

ErikBjorge: Sure, not a problem.  I will push a new patch.

Sae86 (approved):

npmitche (approved):

PR #1305, opened 2021-11-15
Modify is_supported and logic to support other OS

BrentHoltsclaw:

### DIFF
--- a/chipsec/modules/common/wsmt.py
+++ b/chipsec/modules/common/wsmt.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 # CHIPSEC: Platform Security Assessment Framework
 # Copyright (c) 2021, SentinelOne
+# Copyright (c) 2021, Intel Corporation
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


npmitche:
@fishilico 
Looks like the archlinux tests are failing but I don't know why. Nothing changed with respect to the DKMS setup that should cause it to fail. Can you look into it?

fishilico:
> Looks like the archlinux tests are failing but I don't know why. Nothing changed with respect to the DKMS setup that should cause it to fail. Can you look into it?

From a quick look, it seems that DKMS 3.0 was released a few days ago (https://github.com/dell/dkms/releases) and Arch Linux applied the upgrade yesterday (https://github.com/archlinux/svntogit-packages/commits/packages/dkms?after=c4124e2e180e200b816efd019415af789e56055c+0&branch=packages%2Fdkms). Now on Arch Linux this command fails:

```console
$ dkms add drivers/linux
Error! Arguments <module> and <module-version> are not specified.
Usage: add <module>/<module-version> or
       add -m <module>/<module-version> or
       add -m <module> -v <module-version>
```

I will search DKMS documentation about how modules are supposed to be built since the latest release. Thanks for the heads up!

npmitche (approved):

frinzell (approved):

Sae86 (approved):

PR #1322, opened 2021-12-03
Cleanup register_high code for split MMIO bars

### LINKED ISSUE #1317: Remove register_high for split MMIO BARs
The standard approach in chipsec for those is to create a fake register of size 8 spanning the low and high registers defining the BAR (at least if both registers are consecutive, which seems to always be the case for both Intel and AMD).

kerneis-anssi:
Closes #1317.

@headkaze can you check this doesn't break you on AMD by any chance? If we want to lend this for the next release, I'd be more confident with someone else than me testing it. It shouldn't change any behavior, it's a pure cleanup (hopefully).


frinzell (approved):

Sae86 (approved):

kerneis-anssi:
I think it's ready to merge, but I'll let you do because I'm unclear when you'd rather have it (before or after the release cut).

PR #1326, opened 2021-12-08
Modify swsmi asm to break up B2 and B3

BrentHoltsclaw:
We noticed on certain systems that SMI were being triggered and dropped. Digging a little deeper we found that 0xB2 needed to be a BYTE. Please see:
[PchSmmSw.c](https://github.com/tianocore/edk2-platforms/blob/aa3f6fd542e99dde4206537b095f1a2201275e75/Silicon/Intel/CoffeelakeSiliconPkg/Pch/PchSmiDispatcher/Smm/PchSmmSw.c#L314)

 //
    // If the IO address is not "BYTE" "WRITE" to "R_PCH_IO_APM_CNT (0xB2)", skip it.
    //
    if ((SmiIoInfo.IoPort != R_PCH_IO_APM_CNT) ||
        (SmiIoInfo.IoType != EFI_SMM_SAVE_STATE_IO_TYPE_OUTPUT) ||
        (SmiIoInfo.IoWidth != EFI_SMM_SAVE_STATE_IO_WIDTH_UINT8))
    {
      continue;
    }

### DIFF
--- a/drivers/linux/amd64/cpu.asm
+++ b/drivers/linux/amd64/cpu.asm
@@ -419,10 +419,9 @@ __swsmi__:
     mov rdi, r12 ; rdi_value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
-    push rax
-    shr ax, 8
+    ror ax, 8
     out 0B3h, al
-    pop rax
+    ror ax, 8
     out 0B2h, al
 
     mov r12, rdi ; rdi_value
--- a/drivers/linux/amd64/cpu.asm
+++ b/drivers/linux/amd64/cpu.asm
@@ -418,7 +418,8 @@ __swsmi__:
 
     mov rdi, r12 ; rdi_value
 
-    ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
+    ; these OUT instructions will write BYTE value (smi_code_data) to port 0xB3 then port 0xB3 (SW SMI control and data ports)
+    ; the writes need to be broken up as some systems will drop the interrupt if the port size is larger than a BYTE
     ror ax, 8
     out 0B3h, al
     ror ax, 8
--- a/drivers/linux/amd64/cpu.asm
+++ b/drivers/linux/amd64/cpu.asm
@@ -419,7 +419,11 @@ __swsmi__:
     mov rdi, r12 ; rdi_value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
-    out 0B2h, ax
+    push rax
+    shr ax, 8
+    out 0B3h, al
+    pop rax
+    out 0B2h, al
 
     mov r12, rdi ; rdi_value
     pop rdi
--- a/drivers/win7/amd64/cpu.asm
+++ b/drivers/win7/amd64/cpu.asm
@@ -407,7 +407,13 @@ _swsmi PROC FRAME
     xchg rdi, [r10+30h]  ; //rdi value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
-    out 0B2h, ax ; 0xB2
+    push rax
+    .pushreg rax
+    .endprolog
+    shr ax, 8
+    out 0B3h, al ; 0xB3
+    pop rax
+    out 0B2h, al ; 0xB2
 
     ; some SM handlers return data/errorcode in GPRs, need to return this to the caller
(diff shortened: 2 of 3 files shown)


npmitche (approved):

drivers/linux/amd64/cpu.asm
```diff
@@ -419,7 +419,11 @@ __swsmi__:
     mov rdi, r12 ; rdi_value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
-    out 0B2h, ax
+    push rax
```
iherna2, line 422: nit: isn't is the same  to do
`out 0b3h, ah`
`out 0b2h, al`

iherna2 (approved):

drivers/linux/amd64/cpu.asm
minipli-oss: It's more of that there's no instruction encoding for OUT using AH. Only AL, AX and EAX can be used.

drivers/linux/amd64/cpu.asm
```diff
@@ -419,7 +419,11 @@ __swsmi__:
     mov rdi, r12 ; rdi_value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
-    out 0B2h, ax
+    push rax
+    shr ax, 8
+    out 0B3h, al
+    pop rax
+    out 0B2h, al
```
minipli-oss, line 426: This really looks worse that what it was before. Can we instead just add a comment that we're writing both ports at once?

minipli-oss: Ahh, that's actually some useful information to add to the commit log or even to the code?

Anyhow, I've some pending driver cleanups, also to this code (see https://github.com/minipli-oss/chipsec/commit/6f8f7f5777fbd2f7f80273c112f12040d4443ea4). But I can rebase these on top of your commit when it gets merged.

minipli-oss: Maybe we can use ROR instead which would avoid the PUSH, which seems to be wrongly annotated for the Windows version?:
```asm
    ror ax, 8
    out 0B3h, al ; 0xB3
    ror ax, 8
    out 0B2h, al ; 0xB2
```

drivers/win7/amd64/cpu.asm
```diff
@@ -407,7 +407,13 @@ _swsmi PROC FRAME
     xchg rdi, [r10+30h]  ; //rdi value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
-    out 0B2h, ax ; 0xB2
+    push rax
+    .pushreg rax
+    .endprolog
```
minipli-oss, line 412: ` .endprolog` seems to be wrong as this is hardly the end of the prologue. Also there's one before in line 394 already.

BrentHoltsclaw: Do you want me to make the changes or get this merged so you can make the updates?

minipli-oss: Either way works for me. But I'd say, at least the `.endprolog` should be dropped before this gets merged.

drivers/linux/amd64/cpu.asm
ErikBjorge: Using rotate is a much better idea to get rid of the push.

BrentHoltsclaw: changes made thanks for the help

drivers/linux/amd64/cpu.asm
```diff
@@ -419,7 +419,10 @@ __swsmi__:
     mov rdi, r12 ; rdi_value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
```
minipli-oss, line 421: This comment is no longer true, as it's two OUT instructions now.

drivers/win7/amd64/cpu.asm
```diff
@@ -407,7 +407,10 @@ _swsmi PROC FRAME
     xchg rdi, [r10+30h]  ; //rdi value
 
     ; this OUT instruction will write WORD value (smi_code_data) to ports 0xB2 and 0xB3 (SW SMI control and data ports)
```
minipli-oss, line 409: dtio

minipli-oss (commented):
Can you please also integrate the information of https://github.com/chipsec/chipsec/pull/1326#discussion_r778976093, either as a comment in the code or in the commit log?

Beside that, LGTM!

ErikBjorge (approved):

minipli-oss (approved):
LGTM!
Thanks, Brent.

PR #1339, opened 2022-01-05
Dev 2.0.0 logging

BrentHoltsclaw:
Changes to logging within CHIPSEC.  

-Make more use of python logging functionality.
- remove log_*_checked functions
- change checks for if logger.HAL/VERBOSE/DEBUG functionality to use log_verbose/hal/debug

### DIFF
--- a/chipsec_main.py
+++ b/chipsec_main.py
@@ -485,7 +485,6 @@ class ChipsecMain:
                 if self.failfast:
                     raise msg
                 return  ExitCode.EXCEPTION
-                logger().warn("Platform dependent functionality is likely to be incorrect")
             except OsHelperError as os_helper_error:
                 logger().error(str(os_helper_error))
                 logger().log_debug(traceback.format_exc())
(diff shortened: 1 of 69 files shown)


chipsec_main.py
```diff
@@ -488,15 +484,14 @@ def main(self):
...
                 logger().error('To specify a pch please use --pch command-line option\n')
                 logger().error('To load legacy configuration and run anyways please use -i command-line option')
-                if logger().DEBUG:
-                    logger().log_bad(traceback.format_exc())
+                logger().log_debug(traceback.format_exc())
                 if self.failfast:
                     raise msg
-                return ExitCode.EXCEPTION
+                return  ExitCode.EXCEPTION
+                logger().warn("Platform dependent functionality is likely to be incorrect")
```
ErikBjorge, line 488: Seems wrong to return and then want to log something...

ErikBjorge (approved):

PR #1341, opened 2022-01-05
Add routines to check for all 1's (0xFs)

frinzell:
Simple check if a value is all 0xF's (1's) and routines based on register/field definition.

### DIFF
--- a/chipsec/defines.py
+++ b/chipsec/defines.py
@@ -1,5 +1,5 @@
 #CHIPSEC: Platform Security Assessment Framework
-#Copyright (c) 2010-2021, Intel Corporation
+#Copyright (c) 2010-2022, Intel Corporation
 #
 #This program is free software; you can redistribute it and/or
 #modify it under the terms of the GNU General Public License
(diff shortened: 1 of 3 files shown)


chipsec/defines.py
```diff
@@ -197,6 +197,10 @@ def is_printable(seq):
 def is_hex(maybe_hex):
     return all(char in string.hexdigits for char in maybe_hex)
 
+def is_all_ones(value, size, width=8):
```
BrentHoltsclaw, line 200: width=4?

chipsec/defines.py
```diff
@@ -197,6 +197,10 @@ def is_printable(seq):
...
+    mask = (1<<(size * width)) - 1
```
BrentHoltsclaw, line 201: spacing around <<

chipsec/defines.py
frinzell: Register size of 1 is 8 bits as default

npmitche (approved):

Sae86 (approved):

PR #1348, opened 2022-01-24
Dev 2.0.0 lib hal split

BrentHoltsclaw:
Series of commits to move files and functions within hal that do not need cs or helper into a new lib directory


chipsec/hal/spi.py
```diff
@@ -853,3 +853,184 @@ def get_SPI_JEDEC_ID_decoded(self):
         part = JEDEC_ID.DEVICE.get( jedec_id, 'Unknown')
 
         return (jedec_id, manu, part)
+
+    def parse_spi_flash_descriptor(self, rom):
```
ErikBjorge, line 857: I'm not sure about how this code has been refactored into the SPI HAL.  The code the processes the raw ROM data should be put in one or more library functions.  The register accesses should have interfaces added to the SPI HAL.  The spidesc util command should then be updated to call the desired functions and format the output.

ErikBjorge (changes requested):
In general this looks good.  I think that just moving parse_spi_flash_descriptor is not the correct way to refactor the code.  Some of the code is specifically data processing other parts should be moved to the SPI HAL.

ErikBjorge (approved):

PR #1349, opened 2022-01-24
Logging changes 2

BrentHoltsclaw:
Series of commits to fix logging changes


ErikBjorge (changes requested):
I think the file chipsec/lib/uefi_platform.py belongs in the library refactor and not as part of the logging changes.  Move chipsec/lib/uefi_platform.py to the other PR and this one is good.

ErikBjorge (approved):

PR #1361, opened 2022-01-31
Logger updates to better integrate with Python Logging functionality

BrentHoltsclaw:
Update the logger functionality

- Add new logging levels
- Create colors for new logging levels
- Refactor log function to integrate with new levels
- Add log_hal, log_verbose, log_debug functions

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -616,16 +616,18 @@ class Chipset:
         if logger().DEBUG:
             logger().log('[*] Loading device buses..')
         if QUIET_PCI_ENUM:
-            old_hal_state = logger().HAL
-            logger().HAL = False
+            old_log_state = (self.logger.HAL, self.logger.DEBUG, self.logger.VERBOSE)
+            self.logger.HAL, self.logger.DEBUG, self.logger.VERBOSE  = (False, False, False)
+            self.logger.setlevel()
         try:
             enum_devices = self.pci.enumerate_devices()
-        except:
+        except Exception:
             if logger().DEBUG:
                 logger().log('[*] Unable to enumerate PCI devices.')
             enum_devices = []
         if QUIET_PCI_ENUM:
-            logger().HAL = old_hal_state
+            self.logger.HAL, self.logger.DEBUG, self.logger.VERBOSE  = old_log_state
+            self.logger.setlevel()
 
         # store entries dev_fun_vid_did = [list of bus entries]
         for enum_dev in enum_devices:
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -616,9 +616,9 @@ class Chipset:
         if logger().DEBUG:
             logger().log('[*] Loading device buses..')
         if QUIET_PCI_ENUM:
-            old_log_state = (self.logger.HAL, self.logger.DEBUG, self.logger.VERBOSE)
-            self.logger.HAL, self.logger.DEBUG, self.logger.VERBOSE  = (False, False, False)
-            self.logger.setlevel()
+            old_log_state = (logger().HAL, logger().DEBUG, logger().VERBOSE)
+            logger().HAL, logger().DEBUG, logger().VERBOSE  = (False, False, False)
+            logger().setlevel()
         try:
             enum_devices = self.pci.enumerate_devices()
         except Exception:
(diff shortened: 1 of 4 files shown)


chipsec/logger.py
```diff
@@ -140,16 +147,17 @@ def __init__( self ):
...
-        self.verbose = 15
+        pyLogging.addLevelName(19, "verbose")
+        pyLogging.addLevelName(18, "hal")
+        pyLogging.addLevelName(29, "bad")
+        pyLogging.addLevelName(28, "good")
+        pyLogging.addLevelName(39, "header")
         self.logstream = pyLogging.StreamHandler(sys.stdout)
-        self.logstream.setFormatter(ColorLogger()) #applys colorization to output
-        self.rootLogger.addHandler(self.logstream) #adds streamhandler to root logger
+        self.logstream.setFormatter(ColorLogger())  # applys colorization to output
```
iherna2, line 159: Nit: Should be "applies"

chipsec/logger.py
```diff
@@ -140,16 +147,17 @@ def __init__( self ):
...
         self.logfile = None
-        self.debug = pyLogging.DEBUG
-        self.info = pyLogging.INFO
         self.rootLogger = pyLogging.getLogger(__name__)
-        self.rootLogger.setLevel(self.debug)
+        self.rootLogger.setLevel(pyLogging.INFO)
         self.ALWAYS_FLUSH = False
-        pyLogging.addLevelName(15, "verbose")
-        self.verbose = 15
+        pyLogging.addLevelName(19, "verbose")
```
iherna2, line 153: Would it be useful to have a named color level?

I am thinking in situations where printing decoded in multiple colors helps with the readability; in those scenarios saying print_header() feels awkward when the intention is actually print_blue()

iherna2 (approved):

chipsec/logger.py
ErikBjorge: The use of print_header and other methods provide additional formatting as well as applying color to the text.  This keeps the log format consistent between modules.  Arbitrary highlighting of text by a module should be avoided.  If this is needed in the long term the log function should be updated to allow for overriding the text color without impacting the consistent text highlighting provided by the other interfaces.

ErikBjorge (approved):

npmitche (approved):

frinzell:
We should keep `verbose_log()` for now.

chipsec/chipset.py
```diff
@@ -562,15 +562,17 @@ def init_cfg_xml(self, fxml, code, pch_code):
     def init_cfg_bus( self ):
         if logger().DEBUG: logger().log( '[*] Loading device buses..' )
         if QUIET_PCI_ENUM:
-            old_hal_state = logger().HAL
-            logger().HAL = False
+            old_log_state = (self.logger.HAL, self.logger.DEBUG, self.logger.VERBOSE)
```
frinzell, line 565: These are all `logger().x`

Sae86 (changes requested):
Win32helper needs updates too

chipsec/logger.py
```diff
@@ -215,148 +203,108 @@ def flush(self):
...
-
-    def log_information_check( self, text ):  # Duplicate of log_information()
-        """Logs a Information test, an information test"""
-        self.log_information(text)
-
-    def log_not_applicable_check( self, text):  # Duplicate of log_not_applicable()
-        """Logs a Test as Not Applicable"""
-        self.log_not_applicable(text)
-
-    def log_result( self, text ):  # Duplicate of log_good()
```
Sae86, line 276: Deprecated, can be removed

chipsec/logger.py
```diff
@@ -215,148 +203,108 @@ def flush(self):
...
-        self.log(text, pyLogging.CRITICAL)
+        self._log(text, pyLogging.INFO, "BLUE")
 
     def log_important(self, text):
         """Logs an important message."""
         text = "[!] " + text
-        self.log(text, pyLogging.ERROR)
+        self._log(text, pyLogging.INFO, "RED")
+
+    def log_result(self, text):
```
Sae86, line 272: Deprecated, can be removed

BrentHoltsclaw:
> Win and linux helpers need updates too

Would it make more sense to handle those with a separate PR?

Sae86:
> Would it make more sense to handle those with a separate PR?

Yes, but probably we should move the changes to efihelper to that PR as well

npmitche (approved):

PR #1363, opened 2022-02-01
Changes logger functions based on #1361

BrentHoltsclaw:
Depends on #1361 

- Convert logging to use log_hal, log_verbose, log_debug when applicable.
- Convert uses of logger.warn into logger.log_warning
- Convert uses of logger.error into logger.log_error
- Remove use of logger _check functions

### DIFF
--- a/chipsec/chipset.py
+++ b/chipsec/chipset.py
@@ -164,9 +164,9 @@ class Chipset:
             did = (vid_did >> 16) & 0xFFFF
             rid = self.pci.read_byte(0, 0, 0, PCI_HDR_RID_OFF)
         except:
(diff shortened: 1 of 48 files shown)


frinzell:
Just the APIs right now. Plus this will be in smaller doses.

ErikBjorge:
What patch size would you find more acceptable for this type of cleanup?  I'm sure these could be submitted over time with 10 or so files per PR.

BrentHoltsclaw:
Is this an acceptable size or do I need to break up the 3 commits into separate PRs?

frinzell:
Let's pick up during our next sync.

PR #1376, opened 2022-02-03
Clairify uefivar_fuzz.py documentation

frinzell:
- Clarify some (possible) expected behaviors
- Use single 'doc' string
- Minor clean-up

### DIFF
--- a/chipsec/modules/tools/uefi/uefivar_fuzz.py
+++ b/chipsec/modules/tools/uefi/uefivar_fuzz.py
@@ -25,9 +25,6 @@ The module is fuzzing UEFI Variable interface.
 The module is using UEFI SetVariable interface to write new UEFI variables
 to SPI flash NVRAM with randomized name/attributes/GUID/data/size.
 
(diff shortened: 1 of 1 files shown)


chipsec/modules/tools/uefi/uefivar_fuzz.py
```diff
@@ -229,4 +212,4 @@ def run( self, module_argv ):
                 status = self._uefi.delete_EFI_variable(bytestostring(_NAME), str(_GUID))
                 self.logger.log(status)
 
-        return ModuleResult.PASSED
+        return ModuleResult.WARNING
```
npmitche, line 215: Most of the tools return passed at the end of the execution. Is there a specific reason why we're not keeping to that?

frinzell: Considering the system could be in an unknown state, even if the module finished, a Warning, further manual evaluation may be needed, seemed more appropriate.

npmitche (approved):

Sae86 (approved):

PR #1395, opened 2022-03-01
Disable colorized output when piping the output or when NO_COLOR is set

kerneis-anssi:
The former case is useful if the user runs some variant of:
  chipsec_main.py > chipsec.log

The latter case is defined in https://no-color.org/ and simple enough to
implement while we are improving color handling.

Signed-off-by: Gabriel Kerneis <gabriel.kerneis@ssi.gouv.fr>


kerneis-anssi:
I just noticed #1361 by @BrentHoltsclaw which also improves logging. It might make sense to roll this minor change into the other PR instead.

frinzell (approved):

Sae86 (approved):

PR #1398, opened 2022-03-02
Use Black for automatic formatting

kerneis-anssi:
WIP, first experiment as a basis for discussion

Needs some documentation for devs to `pip install pre-commit && pre-commit install` so that code is auto-formatted when they commit.


BrentHoltsclaw (approved):

chipsec/defines.py
```diff
@@ -139,67 +139,99 @@ def scan_single_bit_mask(bit_mask):
...
     version_strs = []
     chipsec_folder = os.path.abspath(chipsec.file.get_main_dir())
-    for fname in sorted([x for x in os.listdir(os.path.join(chipsec_folder, "chipsec")) if x.startswith('VERSION')]):
+    for fname in sorted(
+        [
+            x
+            for x in os.listdir(os.path.join(chipsec_folder, "chipsec"))
+            if x.startswith("VERSION")
+        ]
+    ):
```
ErikBjorge, line 216: This level of formatting seems a bit excessive for comprehensions.  However, it is readable and will just take some time to get used to.

chipsec/chipset.py
```diff
@@ -48,30 +48,32 @@
...
+    MMCFG = "mmcfg"
+    MMIO = "mmio"
+    MSR = "msr"
+    PORTIO = "io"
+    IOBAR = "iobar"
+    MSGBUS = "msgbus"
+    MM_MSGBUS = "mm_msgbus"
+    MEMORY = "memory"
+    IMA = "indirect"
+
```
ErikBjorge, line 63: Any particular reason for the conversion to " from ' or is that just the tool normalizing to a single quote style?

npmitche: Looking at the rest of the file, looks like it's just normalization.

chipsec/defines.py
kerneis-anssi: Agreed. See my other comment about customization.

chipsec/chipset.py
kerneis-anssi: The core philosophy of Black is to have a canonical style and just have people adhere to it, fully automating formatting when you "git commit". So you code as you like, and then it's all normalized in the end. In particular, it doesn't make any decision based on the existing style, and only works on full files, not incrementally. They offer very little customization tweaks, and recommend not using them so that people get used to the "Black style" across various Python projects. However, string-quote normalization is one of the few things that can be disabled: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#command-line-options

Apart from documentation, one other thing I want to improve before submitting this patch is coexistence with flake8. In particular, we need at least this tweak:
https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#flake8

frinzell: The formatting changes seem a bit aggressive.  Will it follow something closer to our existing [.flake8](https://github.com/chipsec/chipsec/blob/main/.flake8) file?

frinzell: Consistency, quote consistency, is good but it largely depends on the use case.  CHIPSEC code has been, in general, moving to using single quotes (').  It would be better aligning to that.
My bigger concern is with the line splitting.  I would rather have a person review these changes than to blindly auto format.  Cases like this could be an indication that a particular piece of code could be written better or more cleanly.  That's the better ROI.  And, agreeing with another comment, it would be an adjustment reading it.  Having spent years reading and maintaining this code, it is more difficult for me to read.  IMO.
We have discussed formatting and style guides in the past but have not come to a single point of authority yet.  I do have a contribution guide planned, which I hope to release, but is still under development.

kerneis-anssi: @frinzell Right. Then it looks to me that YAPF would be much better than Black for Chipsec.  I wonder if the specific issue of line wrapping noticed in this PR could be solved by allowing longer lines, but in general you'd want to use the knobs of YAPF to match your styleguide, and disable it selectively when needed, instead of accepting Black's opinionated approach.

You may want to experiment a bit with YAPF, writing a config file that matches your preferred style and see how it looks once applied to some key files of chipsec. That may be easier that describing the rules to follow in the contribution guide in plain text. Let me know if you need a hand setting up pre-commit checks at some point, I'd be happy to help. I'll close this PR in any case. Thanks for the feedback!

By the way, [pyswarm's contribution guide](https://pyswarms.readthedocs.io/en/development/contributing.html) is an example I really like, short and to the point, if you're looking for some inspiration.

PR #1483, opened 2022-05-25
Correct list numbering in manual

dscott90:


docs/sphinx/index.rst
```diff
@@ -90,16 +90,7 @@ Module & Command Development
     :glob:
     :maxdepth: 1
     :caption: Architecture and Modules
-
+    
     development/*
 
-Contribution and Style Guides
```
Sae86, line 96: Seems like this was removed by mistake. I believe the only change needed are the numbers

Sae86 (approved):

PR #1515, opened 2022-06-28
Update logger.error() in hal modules

frinzell:
Replace deprecated `error()` with `log_error()`

Signed-off-by: Frinzell, Aaron <aaron.frinzell@intel.com>


dscott90 (approved):

npmitche (approved):
good spot on the % formatter in hal/iobar. Thought we got them all already.

PR #1519, opened 2022-06-29
Adding a cache for resolved BAR addresses

iherna2:
When manipulating multiple registers in the same MMIO region, the MMIO
base address is calculated on every register access. This cache helps to
save time by eliminating some redundant PCI accesses.

The cache is an opt-in feature and requires to be explicitly enabled.

Signed-off-by: Ignacio Hernandez <ignacio.hernandez@intel.com>

### DIFF
--- a/chipsec/hal/mmio.py
+++ b/chipsec/hal/mmio.py
@@ -62,6 +62,8 @@ class MMIO(hal_base.HALBase):
 
     def __init__(self, cs):
         super(MMIO, self).__init__(cs)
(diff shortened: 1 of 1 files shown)


npmitche (commented):
Are there times when this would break something? BARs don't generally change, so I could see this being useful as an opt out once it's proven unless there are counter examples.

npmitche (approved):

dscott90 (approved):

Sae86 (approved):

PR #1535, opened 2022-07-13
Fix memory allocation bug in smm_ptr

npmitche:

### DIFF
--- a/chipsec/modules/tools/smm/smm_ptr.py
+++ b/chipsec/modules/tools/smm/smm_ptr.py
@@ -147,7 +147,6 @@ _FILL_VALUE_BYTE  = 0x5A
 _SMI_CODE_DATA    = 0x0
 _MEM_FILL_VALUE   = b'\x11'
 _MEM_FILL_SIZE    = 0x500
-_MAX_ALLOC_PA     = 0xFFFFFFFF
 _DEFAULT_GPRS     = {'rax': _FILL_VALUE_QWORD, 'rbx': _FILL_VALUE_QWORD, 'rcx': _FILL_VALUE_QWORD, 'rdx': _FILL_VALUE_QWORD, 'rsi': _FILL_VALUE_QWORD, 'rdi': _FILL_VALUE_QWORD}
 
 _pth = 'smm_ptr'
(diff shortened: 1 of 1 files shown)


dscott90 (approved):

Sae86 (approved):

minipli-oss:
I think the limitation to allocate memory with a physical address below 4GB was intentional. Why do you remove that limitation, i.e. what's the bug?

npmitche:
Yeah, I think you might be right. I'm still investigating.  This is why it has not been merged in yet. 🤔

minipli-oss:
And what's the bug? Can you point me to an issue or provide some logs?

I fixed allocation errors for smm_ptr in PR #1389 specifically. Maybe I broke something else with these fixes?

npmitche:
Closing due to wrong implementation of fix.

minipli-oss:
Well, my fix made sure we're actually using the correct physical address while the old code was silently truncating the allocated address to 32 bit. The the old code, while "working", is doing the wrong thing -- passing a bogus address to the SMI handler.

Can you provide more verbose logs?

npmitche:
Yeah, you're right. In the "working" dmesg I see this:
```
[  549.384400] auto freeing allocated memory (va = 000000004e05855c, pa = 0x10df89000)
[  549.384402] auto freeing allocated memory (va = 00000000b9095912, pa = 0x10df8b800)
[  657.131309] Chipsec module loaded 
[  657.131313] ** This module exposes hardware & memory access, **
[  657.131313] ** which can effect the secure operation of      **
[  657.131314] ** production systems!! Use for research only!   **
[  657.709046] [chipsec] WARNING: allocated memory (0x10472e000) is not below max_pa (0xffffffff) (ignoring)
[  657.709108] [chipsec] WARNING: allocated memory (0x10472f800) is not below max_pa (0xffffffff) (ignoring)
[  657.711767] [chipsec] > IOCTL_SWSMI
[  657.714356] [chipsec] > IOCTL_SWSMI
```

npmitche:
If I comment out:
```python
if (max_pa > 16 * 1024 * 1024)
			va = kmalloc(NumberOfBytes, GFP_KERNEL | GFP_DMA32);
```
It looks like it works. When I looking at the kmalloc api, I don't see a GFP_DMA32 flag definition, just the GFP_DMA one. 
https://www.kernel.org/doc/htmldocs/kernel-api/API-kmalloc.html

minipli-oss:
Can you provide the log of the non-working version, please?

GFP_DMA32 is supposed to return an address below 4GB while GFP_DMA will allocate from the legacy DMA area below 16MB. However, the latter might be occupied by the kernel itself, so the driver tries to use GFP_DMA32 first. But without logs, hard to tell what happens on your system.

npmitche:
[errorlog1.txt](https://github.com/chipsec/chipsec/files/9201225/errorlog1.txt)

PR #1544, opened 2022-07-27
Fix malloc issue in Linux driver

npmitche:
Fixes issue seen in smm_ptr.py and identified in https://github.com/chipsec/chipsec/pull/1535 

### DIFF
--- a/drivers/linux/chipsec_km.c
+++ b/drivers/linux/chipsec_km.c
@@ -43,8 +43,8 @@ chipsec@intel.com
 #endif
 
 
-#define CHIPSEC_VER_ 		1
-#define CHIPSEC_VER_MINOR	2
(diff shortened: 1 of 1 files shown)


Sae86 (approved):

frinzell (approved):

minipli-oss:
Wow, that's a big commit! Can we please separate such big whitespace changes from real bug fixes, so the latter are easier to grasp? Thanks.

That said, it's still not the right fix, as the DMA zone might be otherwise occupied, as already mentioned in https://github.com/chipsec/chipsec/pull/1535#issuecomment-1196988468. Anyhow, I see that GFP_DMA32 is non-functional, unfortunately.

I'll come up with a fallback for failing GFP_DMA allocations later this week.

PR #1562, opened 2022-08-17
Update tests.yml to include python 3.6 in validation

npmitche:
Update tests.yml to include python 3.6 as that is the lowest version of python we support due to the UEFI Shell support.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -163,6 +163,7 @@ jobs:
           - "3.9"
           - "3.8"
           - "3.7"
+          - "3.6"
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


frinzell (approved):

dscott90 (approved):

.github/workflows/tests.yml
```diff
@@ -134,9 +134,11 @@ jobs:
     strategy:
       matrix:
         python:
+          - 3.10
```
fishilico, line 137: Using `3.10` does not work, because the GitHub Action worker believes this is a float and uses `3.1` (this is why the CI tests failed). This value needs to be quoted, which is what I did in PR https://github.com/chipsec/chipsec/pull/1533.
Now that this other PR is merged, this PR can be updated to only add Python 3.6.

npmitche: Yeah, am going to update this. Thanks for pointing this out though!

PR #1590, opened 2022-10-21
Use CPUID to identify processors instead of PCI Dev ID 

jun2zhou:
Add cpuid printing

Signed-off-by: Zhou Jun,  jun2.zhou@intel.com

### DIFF
--- a/chipsec_main.py
+++ b/chipsec_main.py
@@ -446,7 +446,7 @@ class ChipsecMain:
         ret["OS"] = "{} {} {} {}".format(self._cs.helper.os_system, self._cs.helper.os_release,
                                          self._cs.helper.os_version, self._cs.helper.os_machine)
         ret["Python"] = "Python {}".format(platform.python_version())
-        ret["Platform"] = "{}, VID: {:04X}, DID: {:04X}, RID: {:02X}".format(self._cs.longname, self._cs.vid,
+        ret["Platform"] = "{}, CPUID: {}, VID: {:04X}, DID: {:04X}, RID: {:02X}".format(self._cs.longname, self._cs.get_cpuid(), self._cs.vid,
                                                                              self._cs.did, self._cs.rid)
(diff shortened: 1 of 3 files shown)


npmitche:
Do we want to store the CPUID like we are doing with the VID/DID/RID so we don't have to call it again if we need it for a module or HAL?

chipsec_main.py
```diff
@@ -461,10 +461,12 @@ def log_properties(self):
         logger().log("[CHIPSEC] Python  : {} ({})".format(platform.python_version(),
                                                           "64-bit" if self.PYTHON_64_BITS else "32-bit"))
         logger().log("[CHIPSEC] Helper  : {} ({})".format(*self._cs.helper.helper.get_info()))
-        chip_info = "[CHIPSEC] {:8}: {}\n[CHIPSEC]      VID: {:04X}\n" \
+        chip_info = "[CHIPSEC] {:8}: {}\n[CHIPSEC]      CPUID: {}\n[CHIPSEC]      VID: {:04X}\n" \
```
frinzell, line 464: Align to colon :

jun2zhou: resolved

jun2zhou:
changed to ":" aligned display,

cmds tested:

python3 chipsec_main.py

python3 chipsec_main.py -m misc.cpuidtest

python3 chipsec_util.py cpu info

jun2zhou:
> Do we want to store the CPUID like we are doing with the VID/DID/RID so we don't have to call it again if we need it for a module or HAL?

let me rethink about that.

frinzell:
> > Do we want to store the CPUID like we are doing with the VID/DID/RID so we don't have to call it again if we need it for a module or HAL?
> 
> let me rethink about that.

No, We can address this at a later date.

npmitche (approved):

dscott90 (approved):

PR #1616, opened 2022-12-08
Update logger to use python logging

Sae86:
Signed-off-by: Sara Batllori <sara.batllori@intel.com>

### DIFF
--- a/chipsec/logger.py
+++ b/chipsec/logger.py
@@ -21,17 +21,16 @@
 """
 Logging functions
 """
-import logging as pyLogging
+import logging
 import platform
 import string
 import binascii
 import sys
 import os
 import atexit
-from time import localtime
+from time import localtime, strftime
 from typing import Sequence, Tuple, Dict, List
-
-from chipsec.defines import get_version, get_message, os_version
+from enum import Enum
 from chipsec.testcase import ChipsecResults
 
 try:
@@ -47,9 +46,12 @@ except ImportError:
         has_WConio = False
 
 LOG_PATH = os.path.join(os.getcwd(), "logs")
+if not os.path.exists(LOG_PATH):
+    os.mkdir(LOG_PATH)
 
+LOGGER_NAME = 'CHIPSEC'
 
-class chipsecrecordfactory(pyLogging.LogRecord):
+class chipsecRecordFactory(logging.LogRecord):
     try:
         is_atty = sys.stdout.isatty()
     except AttributeError:
@@ -70,6 +72,7 @@ class chipsecrecordfactory(pyLogging.LogRecord):
             }
 
             def getMessage(self) -> str:
+                if self.name != LOGGER_NAME: return super().getMessage()
                 color = None
                 msg = str(self.msg)
                 if self.args:
@@ -106,6 +109,7 @@ class chipsecrecordfactory(pyLogging.LogRecord):
             }
 
             def getMessage(self) -> str:
+                if self.name != LOGGER_NAME: return super().getMessage()
                 color = None
                 msg = str(self.msg)
                 if self.args:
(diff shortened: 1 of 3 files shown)


BrentHoltsclaw:
It appears the PR is removing the 'hal' logging level.  Can you explain why this is being removed.

chipsec/logger.py
```diff
@@ -125,31 +125,76 @@ def getMessage(self) -> str:
...
-        # Respect https://no-color.org/ convention, and disable colorization
-        # when the output is not a terminal (eg. redirection to a file)
-        pyLogging.setLogRecordFactory(chipsecrecordfactory)  # applies colorization to output
-        self.rootLogger.addHandler(self.logstream)  # adds streamhandler to root logger
+        #---------------------------------------------------------------------------------------#
+        configPath = os.path.join(os.getcwd(), 'chipsec', 'logConfig', profile)
+        logPath = os.path.join(os.getcwd(), 'logs', strftime('%a%b%d%y-%H%M%S'))
+        logging.config.fileConfig(configPath, defaults={'filename': repr(logPath)[1:-1]})
+        logging.addLevelName(15, "VERBOSE")
+        logging.addLevelName(19, "verbose")
```
iherna2, line 152: What is the difference between 'VERBOSE' and 'verbose'

chipsec/logger.py
```diff
@@ -125,31 +125,76 @@ def getMessage(self) -> str:
 class Logger:
     """Class for logging to console, text file, XML."""
 
-    def __init__(self):
+    class level(Enum):
+        DEBUG = 10
+        VERBOSE = 11
+        # HAL = 12
```
iherna2, line 131: Why are we removing hal?

chipsec/logger.py
```diff
@@ -194,121 +239,135 @@ def disable(self):
...
 
-    def log(self, text):
-        """Plain Log message"""
-        self._log(text, pyLogging.INFO, "WHITE")
-
-    def log_verbose(self, text):
+    def log_verbose(self, text):  # Use log('text', self.logger.VERBOSE)
         """Logs a Verbose message"""
-        self._log(text, pyLogging.getLevelName("verbose"), "LIGHT_GRAY")
+        #self._log(text, logging.getLevelName("verbose"), "LIGHT_GRAY")
```
iherna2, line 273: please remove dead code

chipsec/logger.py
```diff
@@ -507,13 +563,13 @@ def print_buffer_bytes(arr, length=16):
 
 def pretty_print_hex_buffer(arr, length=16):
     """Prints the buffer (bytes, bytearray) in a grid"""
-    _str = ["    _"]
+    _str = ['    _']
     for n in range(length):
-        _str += ["{:02X}__".format(n)]
+        _str += ['{:02X}__'.format(n)]
```
iherna2, line 568: shouldn't it be _str += [f'{n:02X}__'] for consistency?

chipsec/logger.py
```diff
@@ -194,121 +247,135 @@ def disable(self):
...
-        self._log(text, pyLogging.getLevelName("verbose"), "LIGHT_GRAY")
+        #self._log(text, logging.getLevelName("verbose"), "LIGHT_GRAY")
+        self.log(text, self.level.VERBOSE)
 
-    def log_hal(self, text):
+    def log_hal(self, text):  # Use log("text", self.logger.VERBOSE)
         """Logs a hal message"""
-        self._log(text, pyLogging.getLevelName("hal"), "LIGHT_GRAY")
+        #self._log(text, logging.getLevelName("hal"), "LIGHT_GRAY")
+        self.log(text, self.level.VERBOSE)
```
BrentHoltsclaw, line 287: self.level.HAL?

chipsec/logger.py
```diff
@@ -125,31 +125,84 @@ def getMessage(self) -> str:
...
+        INFO = 20
+        GOOD = 21
+        IMPORTANT = 22
+        WARNING = 30
+        BAD = 31
+        ERROR = 40
+        CRITICAL = 50
+        EXCEPTION = 60
+
+    def __init__(self, profile = 'debug.conf'):
```
BrentHoltsclaw, line 142: remove whitespace around equals parameter

chipsec/logger.py
```diff
@@ -125,31 +125,84 @@ def getMessage(self) -> str:
...
-        # when the output is not a terminal (eg. redirection to a file)
-        pyLogging.setLogRecordFactory(chipsecrecordfactory)  # applies colorization to output
-        self.rootLogger.addHandler(self.logstream)  # adds streamhandler to root logger
+        self.logstream = logging.StreamHandler(sys.stdout)
+        logging.addLevelName(19, "verbose")
+        logging.addLevelName(18, "hal")
+        #---------------------------------------------------------------------------------------#
+        configPath = os.path.join(os.getcwd(), 'chipsec', 'logConfig', profile)
+        logPath = os.path.join(os.getcwd(), 'logs', strftime('%a%b%d%y-%H%M%S'))
+        logging.config.fileConfig(configPath, defaults={'filename': repr(logPath)[1:-1]})
```
BrentHoltsclaw, line 153: looks like chipsec is missing the logs directory may want to check if it exists and create if it doesn't

chipsec/logger.py
```diff
@@ -194,121 +247,135 @@ def disable(self):
...
         except BaseException:
             print(text)
 
-    def log(self, text):
-        """Plain Log message"""
-        self._log(text, pyLogging.INFO, "WHITE")
-
-    def log_verbose(self, text):
+    def log_verbose(self, text):  # Use log('text', self.logger.VERBOSE)
         """Logs a Verbose message"""
```
BrentHoltsclaw, line 280: Can log_hal, log_verbose, and log_debug be left off the deprecated function list.  I see value in calling them directly.

Sae86: The idea is to only have one log function. Do you not see the same value between log_hal() and log(level.HAL)? What makes those better options?

Sae86: Exactly. We can discuss details later

chipsec/logger.py
Sae86: None other than case, since it is case sensitive. Updated to just use Upper case

chipsec/logger.py
Sae86: Not removing it, it was a placeholder. It's now been added.

chipsec/logger.py
```diff
@@ -345,12 +374,12 @@ def print_banner(self, arguments: Sequence[str]) -> None:
         self.log(f'[CHIPSEC] Arguments: {args}')
         self.log(message)
 
-    def print_banner_properties(self, cs) -> None:
+    def print_banner_properties(self, cs, os_version) -> None:
```
BrentHoltsclaw, line 377: can this be changed to pass in the helper string from calling cs.helper.helper.get_info() opposed to expecting a cs object?

chipsec/logger.py
BrentHoltsclaw: When would be a good time to discuss?

frinzell (approved):

chipsec/logger.py
```diff
@@ -70,6 +71,7 @@ class chipsecrecordfactory(pyLogging.LogRecord):
             }
 
             def getMessage(self) -> str:
+                if self.name != "CHIPSEC": return super().getMessage()
```
npmitche, line 74: See comment below: re "CHIPSEC" string

npmitche: ... comment above... depending on how/where you read this.

chipsec/logger.py
Sae86: let's set up a meeting

dscott90 (approved):

npmitche (approved):

chipsec/logger.py
Sae86: will look into this for next release

PR #1648, opened 2023-01-18
Fix issue with modifying variables in variables.py

npmitche:

### DIFF
--- a/chipsec/modules/common/secureboot/variables.py
+++ b/chipsec/modules/common/secureboot/variables.py
@@ -73,7 +73,7 @@ class variables(BaseModule):
     def can_modify(self, name, guid, data, attrs):
         self.logger.log("    > Attempting to modify variable {}:{}".format(guid, name))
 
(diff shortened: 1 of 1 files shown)


BrentHoltsclaw (approved):

chipsec/modules/common/secureboot/variables.py
```diff
@@ -73,7 +73,7 @@ def is_supported(self):
     def can_modify(self, name, guid, data, attrs):
         self.logger.log("    > Attempting to modify variable {}:{}".format(guid, name))
```
BrentHoltsclaw, line 75: Should there be a check that data length is long enough to be modified?

dscott90 (approved):

chipsec/modules/common/secureboot/variables.py
npmitche: I'll look into it, but for now, this works. =)

PR #1649, opened 2023-01-19
Remove bytestostring call in edk2.writemem

rhelmot:
edk2.writemem accepts a bytestring. Attempting to decode the argument will interpret it as latin-1, and then edk2.writemem will re-encode it as utf-8, leading to data corruption.


npmitche (approved):
Tested and no issues seen.

dscott90 (approved):

frinzell (approved):

PR #1686, opened 2023-02-16
Update werkzeug version in log_parser\requirements.txt

frinzell:


jun2zhou (approved):

dscott90 (approved):

johnloucaides:
We can merge after someone has tested it, but it may be better to have test automation added here:
https://github.com/chipsec/chipsec/blob/main/.github/workflows/tests.yml

frinzell:
@mtnmts you submitted this feature a while back.  Werkzeug 0.16.0 (any version <2.2.3) is associated with a disclosure.  Updating to 2.2.3 appears to require the following updates:

Click==8.0.1
Flask==2.2.3
itsdangerous==2.1.2
Jinja2==3.1.2
lxml>=4.6.5
MarkupSafe==2.1.1
Werkzeug==2.2.3

Do these updates work with your use case?  

I tried running the service and parsing a xml log file within Ubuntu but received a 400 error.  I get the same error using the prior versions.  At least I am getting consistent behavior.  :)  Seems I might be missing an important piece but it's unclear what that might be.

PR #1701, opened 2023-02-23
fix windows kernel api version dependent issue

jun2zhou:
[CHIPSEC] OS      : Windows 10 10.0.18363 AMD64 is not working, chipsec driver can not be loaded correctly in SKL.

https://learn.microsoft.com/en-us/windows-hardware/drivers/gettingstarted/platforms-and-driver-versions

### DIFF
--- a/drivers/win7/driver.c
+++ b/drivers/win7/driver.c
@@ -34,6 +34,16 @@ chipsec@intel.com
 
 #pragma code_seg("chipsec_code$__c")
 
+typedef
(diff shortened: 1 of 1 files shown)


dscott90 (approved):

jun2zhou:
test log on old windows version:
[chipsec_main_log.txt](https://github.com/chipsec/chipsec/files/10830561/chipsec_main_log.txt)

PR #1702, opened 2023-02-23
Move 'limit_field' warning under HAL messaging

frinzell:


dscott90 (approved):

chipsec/hal/mmio.py
```diff
@@ -223,7 +223,8 @@ def get_MMIO_BAR_base_address(self, bar_name: str, bus: Optional[int] = None) ->
                 limit_field = bar['limit_field']
                 limit = self.cs.read_register_field(bar_reg, limit_field, bus=_bus)
             else:
-                self.logger.log_warning(f"[mmio] 'limit_field' field not defined for bar, using limit = 0x{limit:X}")
+                if self.logger.HAL:
```
BrentHoltsclaw, line 226: Since this is being moved to only show under HAL logging, does it still need to be log_warning or should it move to log_hal?

BrentHoltsclaw (approved):

chipsec/hal/mmio.py
frinzell: We have a lot of these throughout the HAL; overall, this is something we are monitoring and hope to minimize.

PR #1707, opened 2023-03-02
search both deprecated and new apis from memory, void compiling error on diff compiler envs

jun2zhou:
memory search both apis:
ExAllocatePool2()
ExAllocatePoolWithTag()


jun2zhou:
to resolve conflict

PR #1739, opened 2023-04-11
f-string updates for cpu, secureboot and smm folders in tools

Sae86:


chipsec/modules/tools/smm/rogue_mmio_bar.py
```diff
@@ -174,7 +174,7 @@ def run(self, module_argv):
 
         self.logger.log("[*] Testing MMIO of PCIe devices:")
         for (b, d, f, _, _) in pcie_devices:
-            self.logger.log("    {:02X}:{:02X}.{:X}".format(b, d, f))
+            self.logger.log(f'    {b:02X}:{d:02X}.{f:X}')
```
npmitche, line 177: We should verify and align on how we print BDF values throughout the framework.

npmitche (approved):

PR #1745, opened 2023-04-14
Update to not require DBX

npmitche:
Per new Microsoft requirements: 
https://learn.microsoft.com/en-us/windows-hardware/drivers/install/preproduction-driver-signing-and-install#provisioning-steps 

Also, update to f'strings and added type hints.


dscott90 (approved):

chipsec/modules/common/secureboot/variables.py
```diff
@@ -63,44 +65,48 @@ def __init__(self):
...
         return supported
 
-    def can_modify(self, name, guid, data, attrs):
-        self.logger.log("    > Attempting to modify variable {}:{}".format(guid, name))
+    def can_modify(self, name: str, guid: Optional[AnyStr], data: Optional[bytes]) -> bool:
+        if not guid or not data:
+            self.logger.log(f'    > Attempting to modify variable {guid}:{name}')
+            return False
+        else:
+            self.logger.log(f'    > Attempting to modify variable {guid}:{name}')
```
frinzell, line 80: This log is duplicated, were they meant to be different?

npmitche: Good catch, thanks!

frinzell (approved):

PR #1750, opened 2023-04-20
Add missing methods to dalhelper.py

frinzell:


chipsec/helper/dal/dalhelper.py
```diff
@@ -189,6 +186,15 @@ def read_phys_mem(self, phys_address_hi: int, phys_address_lo: int, length: int)
     def write_phys_mem(self, phys_address_hi: int, phys_address_lo: int, length: int, buf: bytes) -> int:
         return self.write_physical_mem((phys_address_hi << 32) | phys_address_lo, length, buf)
 
+    def va2pa(self, va):
```
npmitche, line 189: Should we have type hinting on stubbed functions?

npmitche (approved):

chipsec/helper/dal/dalhelper.py
dscott90: I think we should, to maintain uniformity across definitions.

dscott90 (approved):

PR #1752, opened 2023-04-20
Type Hint and f-string updates to linuxhelper.py

frinzell:
- Add Type Hints
- Move to f-strings
- Some code clean-up

### DIFF
--- a/chipsec/helper/linux/linuxhelper.py
+++ b/chipsec/helper/linux/linuxhelper.py
@@ -34,7 +34,9 @@ import resource
 import struct
 import subprocess
 import sys
-
+from typing import Dict, List, Optional, Tuple, Iterable, TYPE_CHECKING
+if TYPE_CHECKING:
+    from chipsec.library.types import EfiVariableType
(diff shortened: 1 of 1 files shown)


chipsec/helper/linux/linuxhelper.py
```diff
@@ -1042,124 +1043,124 @@ def EFIVARS_list_EFI_variables(self):
...
         import multiprocessing
         return multiprocessing.cpu_count()
 
     #
     # Speculation control
     #
-    def retpoline_enabled(self):
-        raise UnimplementedAPIError("retpoline_enabled")
+    def retpoline_enabled(self) -> bool:
+        return False
```
npmitche, line 1163: This should stay as unimplemented. In UEFI Shell, I can see this just returning False. But in Linux it is possible to have it, we just haven't implemented the check for it.

npmitche (changes requested):

npmitche (approved):

Sae86 (approved):

PR #1767, opened 2023-04-26
Add workaround if lzma is not importable.

npmitche:


BrentHoltsclaw:
We may want to put some kind of message that when EfiCompress or LZMA does not import that results may be missing.

Sae86 (approved):

dscott90 (approved):

PR #1777, opened 2023-05-03
Remove filehelper

npmitche:


dscott90 (approved):

BrentHoltsclaw:
May want to look into removing the following pattern throughout the oshelper file as well:
```
        if self.filecmds is not None:
            self.filecmds.AddElement("read_pci_reg", (bus, device, function, address, size), ret)
```

npmitche:
I have that in plan for a future PR specific to cleaning up the oshelper.py.

Sae86 (approved):

PR #1782, opened 2023-05-08
Close file if exception thrown during read

dscott90:


chipsec/file.py
```diff
@@ -46,20 +46,18 @@
 
 def read_file(filename: str, size: int = 0) -> bytes:
     try:
-        f = open(filename, 'rb')
+        with open(filename, 'rb') as f:
+            if size:
+                _file = f.read(size)
+            else:
+                _file = f.read()
+            logger().log_debug(f"[file] Read {len(_file):d} bytes from '{filename:256}'")
```
npmitche, line 54: Recommend adding the `.` in there: `'{filename:.256}'` otherwise it'll pad the file name with white space till it gets to the 256 character mark.
```python
>>> print(f'[file]... from "{fn:256}"')
[file]... from "test.txt

                                "
>>> print(f'[file]... from "{fn:.256}"')
[file]... from "test.txt"
```

npmitche (approved):

Sae86 (approved):

PR #1791, opened 2023-05-10
Remove native api and add linuxnative as separate helper

npmitche:


chipsec/helper/linuxnative/Makefile
```diff
@@ -0,0 +1,2 @@
+clean:
```
BrentHoltsclaw, line 1: I don't think makefile is needed for native

chipsec/helper/linuxnative/linuxnativehelper.py
```diff
@@ -0,0 +1,462 @@
...
+                    value = value & 0xFF
+                elif size == 2:
+                    value = value & 0xFFFF
+                elif size == 4:
+                    value = value & 0xFFFFFFFF
+
+                return value
+            else:
+                byte = b"\xff"
+                return defines.unpack1(byte * size, size)
```
BrentHoltsclaw, line 197: Will this lead to false results?

chipsec/helper/linuxnative/linuxnativehelper.py
```diff
@@ -0,0 +1,462 @@
...
+        if self.devport_available():
+            os.lseek(self.dev_port, io_port, os.SEEK_SET)
+            if 1 == size:
+                fmt = 'B'
+            elif 2 == size:
+                fmt = 'H'
+            elif 4 == size:
+                fmt = 'I'
+            else:
+                raise ValueError("Invalid size")
```
BrentHoltsclaw, line 341: read condition does not have an else case that raises ValueError, suggest making it consistent

BrentHoltsclaw:
May also need to look into:
exceptions.py
```
class UnimplementedNativeAPIError (UnimplementedAPIError):
    def __init__(self, api_name):
        super(UnimplementedNativeAPIError, self).__init__(api_name)
```

acpi.py
`get_ACPI_table_list` and `get_ACPI_table`

chipsec/helper/linuxnative/linuxnativehelper.py
npmitche: Would it be better to raise an exception in this case?

chipsec/helper/linuxnative/linuxnativehelper.py
npmitche: Read condition for io_port (`read_io_port:318`) does have similar condition. Were you referring to the `read_pci_reg` function?

dscott90 (approved):

Sae86 (approved):

PR #1799, opened 2023-05-18
Add typehints and fstrings to cmos_cmd.py

dscott90:


Sae86 (approved):

chipsec/utilcmd/cmos_cmd.py
```diff
@@ -70,27 +70,27 @@ def requires_driver(self):
...
-    def cmos_dump(self):
+    def cmos_dump(self) -> None:
         self.logger.log("[CHIPSEC] Dumping CMOS memory..")
         self._cmos.dump()
 
-    def cmos_readl(self):
+    def cmos_readl(self) -> None:
         val = self._cmos.read_cmos_low(self.offset)
-        self.logger.log("[CHIPSEC] CMOS low byte 0x%X = 0x%X" % (self.offset, val))
+        self.logger.log(f'[CHIPSEC] CMOS low byte 0x{self.offset} = 0x{val}')
```
fishilico, line 79: If I understand the code correctly, this formats integers (`self.offset` and `val`) as decimal, with a hexadecimal prefix. This seems buggy, and the code would likely be instead:
```python
self.logger.log(f'[CHIPSEC] CMOS low byte 0x{self.offset:X} = 0x{val:X}')
```

What do you think?

npmitche: Agreed.

npmitche (changes requested):
See fishilico's comment.

npmitche (approved):

PR #1809, opened 2023-05-25
Add -nl switch to disable autologging

Sae86:
### no switch

- console output
- log file automatically created in logs folder (with date-time log file name)

### -nl

- console output
- no log file created

### -l

- no console output
- no automatic logging
- log file (with specified log file name) created in logs folder

### DIFF
--- a/chipsec/logger.py
+++ b/chipsec/logger.py
@@ -48,8 +48,7 @@ LOG_PATH = os.path.join(os.getcwd(), "logs")
 if not os.path.exists(LOG_PATH):
     os.mkdir(LOG_PATH)
 
-LOGGER_NAME = 'CHIPSEC'
-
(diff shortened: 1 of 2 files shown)


frinzell (approved):

BrentHoltsclaw:
I've found having the file and stream handler both can be helpful at times.  Is there a way that this functionality can remain and be enabled by an option?  I can also see a use case to have the optional file handler be another type of handler such as a socket, http, or even syslog.

dscott90 (approved):

npmitche (approved):

platomav:
Apparently I forgot to actually submit my review of this PR before getting merged. Oups...

Anyway, there are now other tickets which stumbled upon this. But I do believe that the os.path.join should be sanitized when it receives user input.

PR #1812, opened 2023-05-28
Rework legacy PCI helper

fishilico:
When using chipsec.chipset Python module from an unprivileged user, the Python interpreter crashes with a segmentation fault on an `outl` instruction. This is because `iopl(3)` failed and the "legacy PCI" implementation tried to write a value to an I/O port without being allowed to.

Fix this by checking the return value of `iopl`.

While at it, enable using `error` in the C library, to properly report the `EPERM` error.

While working on this code, it appeared that:

- `self.addr` was written twice in `PORTS`' constructor: once for the `inl` implementation and another one for `outl`.
- Everytime `PORTS` or `LEGACY_PCI` objects were created, some memory was allocated (with `mmap.mmap`) to run the helpers and this memory was never freed.
- The global variables `IN_PORT` and `OUT_PORT` were defined as lists but always used as bytes.

Fix these issues by spliting `self.addr` into two variables, by introducing a class property which re-use previous `PORTS` instances (making this class a singleton) and by making `IN_PORT` and `OUT_PORT` directly `bytes`.

While at it, change the names to the usual camel case convention (`PORTS` -> `Ports` and `LEGACY_PCI` -> `LegacyPci`) and add type hints.

### DIFF
--- a/chipsec/helper/linuxnative/legacy_pci.py
+++ b/chipsec/helper/linuxnative/legacy_pci.py
@@ -27,10 +27,12 @@
 
 
 import mmap
+from ctypes import CDLL, CFUNCTYPE, addressof, c_uint16, c_uint32, c_void_p, get_errno
+from typing import Callable
 
-from ctypes import cdll, c_ubyte, CFUNCTYPE, c_uint32, c_uint16, c_void_p, addressof
+from chipsec.exceptions import OsHelperError
 
-IN_PORT = [
+IN_PORT = bytes((
     0x55,                   # push   %rbp
     0x48, 0x89, 0xe5,       # mov    %rsp,%rbp
     0x89, 0xf8,             # mov    %edi,%eax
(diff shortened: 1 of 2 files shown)


chipsec/helper/linuxnative/legacy_pci.py
```diff
@@ -57,49 +59,54 @@
...
-        size = len(opc)
-        code = (c_ubyte * size)(*opc)
-        self.addr = mmap.mmap(-1, mmap.PAGESIZE, prot=mmap.PROT_READ | mmap.PROT_WRITE | mmap.PROT_EXEC)
-        self.addr.write(bytes(code))
+    def __init__(self) -> None:
+        clib = CDLL("libc.so.6", use_errno=True)
+        if clib.iopl(3) == -1:
+            raise OsHelperError("Unable to use I/O ports using iopl", get_errno())
+
+        self.inl_addr = mmap.mmap(-1, mmap.PAGESIZE, flags=mmap.MAP_PRIVATE, prot=mmap.PROT_READ | mmap.PROT_WRITE | mmap.PROT_EXEC)
```
iherna2, line 74: Why is MAP_PRIVATE needed?

iherna2 (approved):

chipsec/helper/linuxnative/legacy_pci.py
fishilico: By default, `mmap.mmap` creates a `mmap.MAP_SHARED` mapping. In this function, there is no need to have a shared mapping, which is why I added `MAP_PRIVATE`.

Also, while debugging other issues with `/proc/self/maps` I stumbled upon this mapping, described there as `/dev/zero (deleted)`. As this is uncommon, I wanted to change it to `MAP_PRIVATE` and did not find any regression while testing.

npmitche (approved):

frinzell (approved):

PR #1817, opened 2023-05-31
Banner.py use cs.cpuid value

frinzell:


chipsec/banner.py
```diff
@@ -64,7 +64,7 @@ def chipsec_banner_properties(cs: Chipset, os_version: Tuple[str, str, str, str]
 [CHIPSEC] Python  : {python_version} ({python_arch})
 [CHIPSEC] Helper  : {helper_name} ({driver_path})
 [CHIPSEC] Platform: {cs.longname}
-[CHIPSEC]    CPUID: {cs.get_cpuid()}
+[CHIPSEC]    CPUID: {cs.cpuid}
```
npmitche, line 67: Should `cs.get_cpuid()` simply return `cs.cpuid` if it's `not None` and if it is `None`, go through the process of trying to read it? 

Using `cs.cpuid`, you're assuming `cs.init()` has been run already, which is probably okay?

npmitche (approved):

Sae86 (approved):

PR #1838, opened 2023-06-02
Implement basehelper.py as an ABC

frinzell:


BrentHoltsclaw:
@frinzell can you elaborate on why an ABC class is necessary

frinzell:
> @frinzell can you elaborate on why an ABC class is necessary

@BrentHoltsclaw to help enforce consistency between helpers

chipsec/helper/basehelper.py
```diff
@@ -68,163 +70,197 @@ def get_info(self):
...
-    def read_mmio_reg(self, phys_address, size):
-        raise NotImplementedError()
+    @abstractmethod
+    def read_mmio_reg(self, phys_address: int, size: int) -> int:
+        pass
 
-    def write_mmio_reg(self, phys_address, size, value):
-        raise NotImplementedError()
+    @abstractmethod
+    def write_mmio_reg(self, phys_address: int, size: int, value: int) -> int:
```
fishilico, line 89: While looking at the implementation of `write_mmio_reg` in the helpers, I am wondering: what is the meaning of the return value? Why is it `int` instead of `None` or `Any`?

For example the Windows helper returns bytes from `_ioctl`, in https://github.com/chipsec/chipsec/blob/a01d537a758d1f7eca23d3a30b59f0574c552f1e/chipsec/helper/windows/windowshelper.py#L580-L581

The Linux helper does not return anything, in https://github.com/chipsec/chipsec/blob/a01d537a758d1f7eca23d3a30b59f0574c552f1e/chipsec/helper/linux/linuxhelper.py#L393-L395

frinzell: This is basically an example of the misalignment: **linuxnativehelper**, **efihelper**, and **dalhelper** do return an `int`.

npmitche (approved):

chipsec/helper/basehelper.py
npmitche: We should ask the question, "Should `set_*` return a value?" If it's returning a status, shouldn't it just raise an error on failure?

Sae86 (approved):

Issue #1827, opened 2023-06-01
MMIO.write_mmcfg_reg is broken since 1.11.0 because of bar_size


fishilico:
Hello,

Since Pull Request https://github.com/chipsec/chipsec/pull/1803, there is a problem in `chipsec/hal/mmio.py`: function `write_MMIO_reg` has been modified from https://github.com/chipsec/chipsec/blob/0016b0c14c95ecc961723f75267cd5ed7ddbcb49/chipsec/hal/mmio.py#L112 to https://github.com/chipsec/chipsec/blob/745ae9084e777915f672ebc17c8a8fedd2cfddf5/chipsec/hal/mmio.py#L106 (parameter `bar_size` disappeared), but all callers have not been updated. `write_mmcfg_reg` still uses this extra parameter: https://github.com/chipsec/chipsec/blob/a61dbc8d2b9d316a4f646e098200e39305fb5a2f/chipsec/hal/mmio.py#L460

So `MMIO.write_mmcfg_reg` is likely to broken in the latest release (v1.11.0), even though I have not confirmed it.

While looking at this bug, I am thinking that using the BAR size seems to be a nice way to prevent out-of-bound access to the registers and it seems it would be nice to verify that `pciexbar_off + size <= pciexbar_sz` in `write_mmcfg_reg`. But on my system, `cs.mmio.get_MMCFG_base_address()` returns `(pciexbar=0xe0000000, pciexbar_sz=0x1000)` so the size seems too low to be useful. If this is general, this means that `pciexbar_sz` should be ignored in `MMIO.write_mmcfg_reg` and a simple fix would be:

```diff
--- a/chipsec/hal/mmio.py
+++ b/chipsec/hal/mmio.py
@@ -438,9 +438,9 @@ class MMIO(hal_base.HALBase):
         return bar_base, bar_size
 
     def read_mmcfg_reg(self, bus: int, dev: int, fun: int, off: int, size: int) -> int:
-        pciexbar, pciexbar_sz = self.get_MMCFG_base_address()
+        pciexbar, _ = self.get_MMCFG_base_address()
         pciexbar_off = (bus * 32 * 8 + dev * 8 + fun) * 0x1000 + off
-        value = self.read_MMIO_reg(pciexbar, pciexbar_off, size, pciexbar_sz)
+        value = self.read_MMIO_reg(pciexbar, pciexbar_off, size)
[...]

PR #1873, opened 2023-06-22
Fix read/write_mmio_reg call with wrong number of arguments

npmitche:
Fixes https://github.com/chipsec/chipsec/security/code-scanning/663
Fixes https://github.com/chipsec/chipsec/issues/1827

### DIFF
--- a/chipsec/hal/mmio.py
+++ b/chipsec/hal/mmio.py
@@ -438,9 +438,9 @@ class MMIO(hal_base.HALBase):
         return bar_base, bar_size
 
     def read_mmcfg_reg(self, bus: int, dev: int, fun: int, off: int, size: int) -> int:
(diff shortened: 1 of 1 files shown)


Sae86 (approved):

fishilico:
This fixes actually a bug that I reported in https://github.com/chipsec/chipsec/issues/1827 .

Moreover, there could be a similar issue with `read_mmcfg_reg` (the code still works because `read_MMIO_reg` is still taking an optional parameter `bar_size`... which is never used).

npmitche:
Ah, thank you for pointing this out. I have not done a good job of keeping up with the issues as it turns out. 

I agree with your assessment and I'll update the read call as well. Thanks!

fishilico (approved):

dscott90 (approved):

PR #1874, opened 2023-06-23
Add return codes changes to module_common and testcase

Sae86:
plus remap support

Requires #1872

### DIFF
--- a/chipsec/module_common.py
+++ b/chipsec/module_common.py
@@ -21,13 +21,20 @@
 
 """
 Common include file for modules
+
+
 """
 
-import chipsec.logger
 import chipsec.chipset
+from enum import Enum
+from chipsec.defines import bit, is_set
+from chipsec.logger import logger
 
 
 class ModuleResult:
+    # -------------------------------------------------------
+    # Legacy results
+    # -------------------------------------------------------
     FAILED = 0
     PASSED = 1
     WARNING = 2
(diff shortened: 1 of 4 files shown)


npmitche (approved):

chipsec/module_common.py
```diff
@@ -125,4 +180,4 @@ def run(self, module_argv):
 #
 # Common module command line options
 #
-OPT_MODIFY = 'modify'
+OPT_MODIFY = 'modify'
```
BrentHoltsclaw, line 183: missing newline

chipsec/module_common.py
```diff
@@ -36,8 +43,58 @@ class ModuleResult:
     INFORMATION = 5
     NOTAPPLICABLE = 6
     ERROR = -1
-
-
+    # -------------------------------------------------------
+    class status(Enum):
```
BrentHoltsclaw, line 47: Question: will the status be exclusive (single bit set) or can multiple bits be set.

BrentHoltsclaw, line 47: Question: Is there a link to some modules that have been converted to get a better sense of how the capability will work?

Sae86: Multiple bits can be set

Sae86: This PR includes remap.py with support

BrentHoltsclaw, line 47: What about inconsistent settings where only some of the bits are set correctly?

Sae86: There's lots of bits available still, we can have one or more for that case if needed. There'll probably be more added as tests are being modified to add support.

chipsec/module_common.py
npmitche: Is it bad to have no new line at the end of the file?

chipsec/module_common.py
BrentHoltsclaw: Shouldn't that work be done first.  Going through the various modules to see potential areas that need to be highlighted so the list is comprehensive before editing them partially and retrofitting if more are added?

chipsec/module_common.py
BrentHoltsclaw: It will not cause any functional issues but is a common coding practice to help ensure the file ends with the proper CRLF/LF and for adding new functionality/concatenation of files without changing the meaning of the code.

chipsec/module_common.py
Sae86: It was done first, that's where that list came from. Hopefully nothing was missed but it's open for change/upgrade/improvement as modules are modified.

dscott90 (approved):

PR #1900, opened 2023-07-28
Add workaround for config loading bug

npmitche:


dscott90 (approved):

chipsec/cfg/parsers/core_parsers.py
```diff
@@ -149,8 +149,11 @@ def _process_pci_dev(self, vid_str, dev_name, dev_attr):
 
     def _add_dev(self, vid_str, name, pci_info, dev_attr):
         if pci_info:
-            self.cfg.BUS[name] = pci_info['bus']
-            pci_info['bus'] = pci_info['bus'][0]
+            if type(pci_info['bus']) == int: # TODO: Investigate why this is needed
```
frinzell, line 152: `if type(pci_info['bus']) is int:` might be better.

npmitche: Updated.

frinzell (approved):

PR #1904, opened 2023-08-01
Change core_parsers to f'strings and hide register size errors

npmitche:


chipsec/cfg/parsers/core_parsers.py
```diff
@@ -170,7 +170,7 @@ def handle_pci(self, et_node, stage_data):
                 continue
             dev_name = dev_attr['name']
             self._process_pci_dev(stage_data.vid_str, dev_name, dev_attr)
-            self.logger.log_debug('    + {:16}: {}'.format(dev_attr['name'], dev_attr))
+            self.logger.log_debug(f"    + {dev_attr['name']:16}: {dev_attr}")
```
BrentHoltsclaw, line 173: should it be single quote?

BrentHoltsclaw (approved):

chipsec/cfg/parsers/core_parsers.py
frinzell: Not with single quote `'name'`

jun2zhou (approved):

PR #1907, opened 2023-08-03
Fix helper record test

jun2zhou:
comparing test context by lines, ignore the Unix & window line ending difference.

### DIFF
--- a/tests/helpers/test_recordhelper.py
+++ b/tests/helpers/test_recordhelper.py
@@ -18,8 +18,6 @@
 
 
 # To execute: python -m unittest tests.helpers.test_recordhelper
-
-from filecmp import cmp
 from os import remove
 import unittest
 from os.path import join, isfile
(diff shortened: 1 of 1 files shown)


tests/helpers/test_recordhelper.py
```diff
@@ -98,6 +96,21 @@ def test_record(self):
 
         self.assertTrue(self.recordhelper.stop())
         self.assertTrue(self.recordhelper.delete())
+        self.assertTrue(self._compare_replay_and_record())
+
+    def _compare_replay_and_record(self):
+        file1 = open(self.original_file, 'r')
```
npmitche, line 102: Would prefer:
```python
with open(self.original_file, 'r') as of:
    original_lines = of.readlines()
with open(self.temp_file, 'r') as tf:
    temp_lines = tf.readlines()
```

npmitche (approved):

tests/helpers/test_recordhelper.py
jun2zhou: will change, thanks your comments.

jun2zhou: done

dscott90 (approved):

PR #1944, opened 2023-09-22
update decoded reporting of SMBIOS Type 0 table

WilliamLeara:
SMBIOS module never updated to report Type 0 fields beyond those found in SMBIOS v2.0 (1996?!?!?).  Add decoding for all fields as of SMBIOS v3.6.0.

### DIFF
--- a/chipsec/hal/smbios.py
+++ b/chipsec/hal/smbios.py
@@ -114,15 +114,15 @@ SMBIOS_STRUCT_TERM_VAL = 0x0000
 
 
 SMBIOS_BIOS_INFO_ENTRY_ID = 0
-SMBIOS_BIOS_INFO_2_0_ENTRY_FMT = '=BBHBBHBBQ'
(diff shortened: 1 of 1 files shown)


npmitche (approved):

npmitche:
Does it make sense to try and read the smbios version first before assuming v3.2+?

dscott90 (approved):

jun2zhou (approved):

WilliamLeara:
> Does it make sense to try and read the smbios version first before assuming v3.2+?

It's a good question.  One approach would be to check SMBIOS version, and then only show those fields introduced/supported in that version; this could get hairy if a particular table were updated (for example) in v2.8, and then v3.1, and then 3.3, etc.  Lots of conditionals.

The other approach would be to assume compliance to the latest spec, but the behavior is undefined if new fields are not populated.

I think the former approach would be more robust, but at the cost of complexity, however it's probably the best way to go.  What do you all think?

npmitche:
I think the former is probably the better direction. 

@frinzell Do you have a preference?

WilliamLeara:
We're in synch.  Let me reject this PR and re-do it.  Thanks all.

PR #1949, opened 2023-09-24
Add return codes support to vmm.venom and wsmt in tools

Sae86:


chipsec/modules/tools/wsmt.py
```diff
@@ -42,6 +42,9 @@
 
 
 class wsmt(BaseModule):
+    def __init__(self):
```
npmitche, line 45: Should not have two __init__ functions.

npmitche (changes requested):

npmitche (approved):

frinzell (approved):

PR #1963, opened 2023-10-17
Update sgx_check unit test and add script to shorten recordings

npmitche:
Attempting to stream line unit testing modules. Using sgx_check as test vehicle.


scripts/strip_record_json_of_pcienumeration.py
```diff
@@ -0,0 +1,95 @@
...
+        self.enum_data = {}
+        self.set_filename(filename)
+        if self.filename:
+            self.load_data()
+        
+    def set_filename(self, filename: str) -> None:
+        self.filename = filename.rstrip(".json")
+        
+    def load_data(self) -> None:
+        if not self.filename:
```
dscott90, line 39: How do you feel about this check being inside set_filename instead?

npmitche: I wanted to leave it for the load_data in case you use these functions in a separate function and you want to load the filename after the initialization of the class. In that case I only want to bail if you try to load_data without a file. I don't care before that.

dscott90 (approved):

Sae86 (approved):

PR #1965, opened 2023-10-23
Add tpm_cmd and txt_cmd unit testing

Sae86:


chipsec/utilcmd/txt_cmd.py
```diff
@@ -162,5 +163,10 @@ def txt_state(self):
         self._log_register("TXT_PCH_DIDVID")
         self._log_register("INSMM")
 
+    def run(self):
+        try:
+            self.func()
+        except Exception:
+            return ExitCode.ERROR
```
npmitche, line 170: Instead of returning ExitCode.Error, set self.ExitCode = ExitCode.Error. This should fix the "Explicit returns mixed with implicit returns" error.
```suggestion
            self.ExitCode = ExitCode.ERROR
```

chipsec/utilcmd/tpm_cmd.py
```diff
@@ -88,15 +89,18 @@ def tpm_state(self):
...
                 self.logger.log(msg)
                 return
 
-        self.func()
-
+    def run(self):
+        try:
+            self.func()
+        except Exception:
+            return ExitCode.ERROR
```
npmitche, line 104: Instead of returning ExitCode.Error, set self.ExitCode = ExitCode.Error. This should fix the "Explicit returns mixed with implicit returns" error.
```suggestion
            self.ExitCode = ExitCode.ERROR
```

npmitche (approved):
LGTM!

dscott90 (approved):

PR #1982, opened 2023-11-09
Add typehints and fstrings to common modules

dscott90:


npmitche (approved):

chipsec/modules/common/rtclock.py
```diff
@@ -121,7 +121,7 @@ def check_rtclock(self):
 
         return self.rc_res.getReturnCode(res)
 
-    def run(self, module_argv):
+    def run(self, *module_argv: str) -> int:
```
npmitche, line 124: This I think is causing tests to fail. All of these should be:
```suggestion
    def run(self, module_argv: List[str]) -> int:
```
I think

npmitche (changes requested):
The run function's args seems like it needs to remove the `*` and add `List[str]` instead of `str`. Needs some more testing.

npmitche (approved):

jun2zhou (approved):

PR #1983, opened 2023-11-10
Add default yaml config options

jun2zhou:


chipsec/utilcmd/smbios_cmd.py
```diff
@@ -32,20 +32,27 @@
...
     def requirements(self) -> toLoad:
         return toLoad.All
 
     def parse_arguments(self) -> None:
+        options = Options()
+        util_cfg = options.get_section_data('Util_Config')
+        try:
+            default_type = util_cfg['smbios_get_type']
+        except:
+            default_type = None
```
npmitche, line 48: Same with this. This should default to "raw"

chipsec/options.py
```diff
@@ -0,0 +1,53 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+#
+# Contact information:
+# chipsec@intel.com
+#
+
+import os
+import yaml
```
npmitche, line 22: Open question: Do we want to use YAML and introduce a new dependency or do we want to use JSON (or similar) which we already use other places and don't need to introduce a new dependency?  
@BrentHoltsclaw  @frinzell

BrentHoltsclaw: YAML is much easier to read and edit than JSON.  If there is a concern about adding a dependency then should probably look into using a configuration file with configparser

frinzell: Small YAML files are fine for reading, larger ones not so much.  The indention-based nesting is the worst, IMO.  Would rather use TOML.  But JSON or even XML would be preferred over YAML.

PR #2075, opened 2024-03-07
Return status in all write port functions

### LINKED ISSUE #625: Empty return statements in io write
I was going through  chipsec/chipsec/hal/io.py and came across empty return-statements in write_port_byte/word/dword methods, though the _write_port method returns status.
I found it a bit inconsistent, therefore, I would like to know whether it was overlooked and the status should be returned in all of the cases. 

```
def _write_port(self, io_port, value, size ):
        if logger().HAL: logger().log( "[io] OUT 0x%04X: value = 0x%08X, size = 0x%02x" % (io_port, value, size) )
        status = self.helper.write_io_port( io_port, value, size )
        return status
```

```
    def write_port_byte(self, io_port, value ):
        if logger().HAL:
            logger().log( "[io] writing byte to I/O port 0x%04X <- 0x%02X" % (io_port, value) )
        self.helper.write_io_port( io_port, value, 1 )
        return
```

Sae86:
Fixes #625

### DIFF
--- a/chipsec/hal/io.py
+++ b/chipsec/hal/io.py
@@ -71,23 +71,23 @@ class PortIO:
             logger().log(f"[io] reading byte from I/O port 0x{io_port:04X} -> 0x{value:02X}")
         return value
 
-    def write_port_byte(self, io_port: int, value: int) -> None:
+    def write_port_byte(self, io_port: int, value: int) -> int:
         if logger().HAL:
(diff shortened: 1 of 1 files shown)


dscott90 (approved):

npmitche (approved):

npmitche (changes requested):
Please change the type hint return value from None.

npmitche (approved):

PR #2090, opened 2024-03-14
Add csv log option and fix value returned after running only one test

Sae86:


dscott90 (approved):

chipsec/logger.py
```diff
@@ -262,6 +265,16 @@ def set_autolog_file(self):
         else:
             print('Unable to autolog')
 
+    def log_csv(self, file_name, test_cases):
+        fields = ['name', 'result', 'code', 'output']
+        if '.csv' not in file_name:
```
npmitche, line 270: should probably be `if not file_name.endswith('.csv'):` That will make sure the .csv is at the end of the string.

chipsec/logger.py
```diff
@@ -262,6 +265,16 @@ def set_autolog_file(self):
...
+            file_name = file_name + '.csv'
+        with open(file_name, 'w') as csv_file:
+            results_csv = csv.DictWriter(csv_file, fieldnames=fields)
+            results_csv.writeheader()
+            for test_case in test_cases:
+                results_csv.writerow(test_case.get_fields())
```
npmitche, line 276: If you wanted to omit the "output" field, you could do: `results_csv.writerow(test_case.get_fields()[:3])` I'll leave it up to you to decide if you want to keep it or night. Doesn't make a big difference.

npmitche (approved):

chipsec/logger.py
Sae86: Doesn't work with dictionaries :(

npmitche (changes requested):
Need to look at unit test failures.

npmitche (approved):

PR #2128, opened 2024-04-12
Add feature to auto-generate/save return code IDs

Sae86:


chipsec/library/returncode.py
```diff
@@ -105,6 +102,10 @@ def getReturnCode(self, result: int) -> int:
         self.resetReturnCodeValues()
         return ret_value
 
+def generate_hash_id(className: str) -> str:
```
npmitche, line 105: Should update to return int:
```suggestion
def generate_hash_id(className: str) -> int:
```

chipsec/module.py
```diff
@@ -72,29 +75,46 @@ def do_import(self):
                     self.logger.log_bad(traceback.format_exc())
                 raise msg
         return loaded
-
+    
+    def get_module_ids_dictionary(self):
+        with open(os.path.join(os.getcwd(), 'chipsec', 'library', 'module_ids.pkl'), 'rb') as module_ids_file:
+            module_ids = pickle.load(module_ids_file)
```
npmitche, line 81: Did you confirm pickle can be used in the shell?

Sae86: yes!

npmitche (approved):

dscott90 (approved):

PR #2169, opened 2024-05-15
Add feature to automatically generate url for return codes

Sae86:
and change module IDs files from pickle to json


chipsec/module.py
```diff
@@ -94,6 +94,17 @@ def get_module_id(self, module_name):
...
         return module_id
     
+    def get_url_info(self):
+        with open(os.path.join(os.getcwd(), 'chipsec', 'library', 'url_format.json'), 'r') as url_file:
+            url_info = json.loads(url_file.read())
+        return url_info
+
+    def get_module_url(self, module_name: str):
+        url_info = self.get_url_info()
+        module_name.replace(url_info['replace'], url_info['with'])
```
npmitche, line 104: Can we change these to something like:
```python
if 'base_url' not in url_info:
    raise exception("Missing Base URL in url file")
replace_find = url_info.get('replace', '')
replace_with = url_info.get('with', '')
endswith = url_info.get('endswith', '')
```
That will give give it a default value if any of these values are not present in the file.

npmitche (approved):
LGTM!

dscott90 (approved):

PR #2200, opened 2024-06-17
smm_ptr: Add SMI contention check to the scan mode

cpey:
Adds a check to the scan mode to ensure there is no SMI contention, which would result in slow SMI execution. It does this by verifying that the MSR value for the SMI count increases by one after each SMI execution.

### DIFF
--- a/chipsec/modules/tools/smm/smm_ptr.py
+++ b/chipsec/modules/tools/smm/smm_ptr.py
@@ -81,6 +81,7 @@ Examples:
 
 import struct
 import os
+import sys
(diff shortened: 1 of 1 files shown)


npmitche:
Any linux specific code should probably go into the linux helper and have the module ask the helper if it either (1) is the right helper or (2) has that function. That or at least wrap the `get_smi_count()` calls in a try/except block?

npmitche (approved):

Sae86 (approved):

PR #2205, opened 2024-06-19
Fix Bios Top Swap Test

jun2zhou:


chipsec/cfg/8086/common.xml
```diff
@@ -690,7 +690,7 @@ Common (default) XML platform configuration file
     <control name="BiosLockEnable"         register="BC"                   field="BLE"                desc="BIOS Lock Enable"/>
     <control name="BiosWriteEnable"        register="BC"                   field="BIOSWE"             desc="BIOS Write Enable"/>
     <control name="TopSwapStatus"          register="BC"                   field="TSS"                desc="Top Swap Status"/>
-    <control name="TopSwap"                register="BUC"                  field="TS"                 desc="Top Swap"/>
+    <control name="TopSwap"                register="BC"                  field="TS"                 desc="Top Swap"/>
```
dscott90, line 693:
```suggestion
    <control name="TopSwap"                register="BC"                   field="TS"                 desc="Top Swap"/>

```

dscott90 (approved):

npmitche (approved):

npmitche (changes requested):
If the TS field is in the BC, the BC needs to be updated to include it. The BC in common.xml does not have it defined.
If this is for a specific platform, should modify the control there, not here.

PR #2206, opened 2024-06-20
Update mmio to check if Base Address is valid

npmitche:
Had to update modules to make CI pass as they are run in VMs and VMs don't have access to hardware, so MMIO BAs are FFs.

### DIFF
--- a/chipsec/hal/mmio.py
+++ b/chipsec/hal/mmio.py
@@ -22,32 +22,32 @@
 Access to MMIO (Memory Mapped IO) BARs and Memory-Mapped PCI Configuration Space (MMCFG)
 
 usage:
-    >>> read_MMIO_reg(cs, bar_base, 0x0, 4)
-    >>> write_MMIO_reg(cs, bar_base, 0x0, 0xFFFFFFFF, 4)
-    >>> read_MMIO(cs, bar_base, 0x1000)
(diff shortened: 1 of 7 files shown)


chipsec/hal/mmio.py
```diff
@@ -22,32 +22,32 @@
...
+    >>> get_MMCFG_base_address()
+    >>> read_mmcfg_reg(0, 0, 0, 0x10, 4)
+    >>> read_mmcfg_reg(0, 0, 0, 0x10, 4, 0xFFFFFFFF)
 """
 from typing import List, Optional, Tuple
 from chipsec.hal import hal_base
 from chipsec.library.exceptions import CSReadError
 from chipsec.library.logger import logger
-from chipsec.library.defines import get_bits
+from chipsec.library.defines import get_bits, is_all_ones
```
dscott90, line 50: Is this supposed to be **is_field_all_ones** to match line 248?

dscott90 (changes requested):

chipsec/hal/mmio.py
npmitche: I don't believe so, is_field_all_ones is on line 215 and it's being called on the register object, not imported.

dscott90 (approved):

Sae86 (approved):

PR #2239, opened 2024-07-25
Fix bug when msgbus_cmd returns value

npmitche:


elicn:
While this PR fixes the problem, it addresses the symptom rather than the root cause. As far as I can tell from the code and function annotations along the way, is that the returned value is of an unexpected class. IMHO it worth to address the root cause by making sure the helper always return the expected type of value.

npmitche:
Fair point. Are you able to test my changes?

elicn:
It works fine here, thanks.

One petty comment though: you don't have to write "0x" every time you want a hexadecimal value prefixed; you can have the prefix emitted as part of the format specifier with `#`. Means, instead of `"0x{self.port:02X}"` you can write `"{self.port:#04X}"` (note the width is +2 now to include the "0x" prefix)

Sae86 (approved):

npmitche:
Pushed update to fix unit test.

dscott90 (approved):

PR #2333, opened 2024-10-16
Add SMI detection accuracy improvements

cpey:
Improves SMI detection with the following changes:

- Flushes cache and enforces memory ordering before triggering SMI.
- Disables interrupt delivery during SMI timing measurement.
- Excludes fast outliers from the average calculation.
- Adds a 10 ms delay before retrying to evade undesired system instability events.

### DIFF
--- a/chipsec/modules/tools/smm/smm_ptr.py
+++ b/chipsec/modules/tools/smm/smm_ptr.py
@@ -82,6 +82,7 @@ Examples:
 import struct
 import os
 import sys
+import time
 
 from chipsec.module_common import BaseModule
 from chipsec.library.returncode import ModuleResult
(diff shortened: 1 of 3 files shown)


chipsec/modules/tools/smm/smm_ptr.py
```diff
@@ -487,6 +506,7 @@ def smi_fuzz_iter(self, thread_id, _addr, _smi_desc, fill_contents=True, restore
             # Re-do the call if it was identified as an outlier, due to periodic SMI delays
             #
             if scan.is_outlier(duration):
+                time.sleep(SCAN_MODE_RETRY_DELAY)
```
npmitche, line 509: I think you're missing the time import, unless I missed it somewhere?

npmitche (approved):

dscott90 (approved):

PR #2343, opened 2024-10-29
Initial submission of AMD common modules

coudrant:
Code created for AMD Strix SOC.  Tested on Ubuntu, Windows and UEFI.  Includes new HAL modules for AMD PSP and changes to hal/cpu.py to support AMD.


npmitche (changes requested):
For files that you are adding to, you are welcome to include your copyright to the top of the files. (See chipsec/command.py for an example)

chipsec/cfg/1022/strix.xml
```diff
@@ -0,0 +1,242 @@
+<?xml version="1.0"?>
+<configuration platform="Renoir">
```
npmitche, line 2: The platform attribute on the configuration tag needs to match the code attributes on the sku tag.

chipsec/modules/common/cpu/cpu_info.py
```diff
@@ -99,4 +100,65 @@ def run(self, module_argv: List[str]) -> int:
...
+        #return self.result.getReturnCode(ModuleResult.INFORMATION)
+        self.res = ModuleResult.INFORMATION
+        return self.res 
+
+    def run(self, module_argv: List[str]) -> int:
+        # Log the start of the test
+        self.logger.start_test('Current Processor Information:')
 
+        if self.cs.is_intel():
+            return self.run_intel()
```
npmitche, line 162: need to pass in `module_argv`: `return self.run_intl(module_argv)`

npmitche (changes requested):
Found 2 more.

chipsec/cfg/1022/strix.xml
```diff
@@ -0,0 +1,242 @@
...
+- PMC events: no support in chipsec for performance monitor counters.
+-->
+
+  <!-- #################################### -->
+  <!--                                      -->
+  <!-- Information                          -->
+  <!--                                      -->
+  <!-- #################################### -->
+  <info family="core">
+    <sku did="0x1507" name="Strix" code="Strix" longname="Strix Point Root Complex" />
```
kerneis-anssi, line 21: I used `did` to identify sku when initially porting AMD to chipsec, but I think it would be better to use `cpuid`.  

Unlike Intel, AMD doesn't identify architecture is terms of PCH device id, but based on family/model/revision. Eg, the Dali PPR is titled "AMD Family 17h Model 20h, Revision A1". This maps to a cpuid of 0820F01 (unless I got the math wrong). I'm not sure what the correct cpuid is for Strix because the PPR isn't public yet, but I'd suggest a line similar to:

    <sku cpuid="0x0820F01" name="dali" code="dali" longname="dali" />

chipsec/cfg/1022/strix.xml
```diff
@@ -0,0 +1,242 @@
+<?xml version="1.0"?>
+<configuration platform="Strix Point">
+<!--
+XML configuration for Strix Point based platforms
+
+Reference: PPR for AMD Family 17h Model 18h B1. 55570-B1 Rev 3.16 - Apr 14, 2021.
```
kerneis-anssi, line 6: Delete this line

chipsec/hal/psp.py
```diff
@@ -0,0 +1,119 @@
...
+import itertools
+import time
+from typing import List, Tuple, Optional
+from chipsec.library.logger import logger, pretty_print_hex_buffer
+from chipsec.library.file import write_file
+from chipsec.library.exceptions import OsHelperError
+from chipsec.library.defines import is_all_ones, MASK_16b, MASK_32b, MASK_64b, BOUNDARY_4KB
+
+class PSP:
+    SMU_INDEX_ADDR = 0xb8
```
superm1, line 35: It's really not nice to go around the kernels back with these mailboxes.

There is a debugfs interface on the mailing list that has proper locking protection.

https://lore.kernel.org/all/20241023172150.659002-17-yazen.ghannam@amd.com/

kerneis-anssi: This is HAL code, which will also be used on Windows and UEFI. It's hardly the only part where Chipsec modules do unreliable accesses that could race with the kernel I believe (eg. I/O ports) so I don't think it should be blocker.

superm1: I see.  Well FYI this type of access will taint the kernel in the future, and eventually want to block it to force moving to the debugfs interface.

superm1: And BTW this naming is totally wrong, this isn't a PSP or SMU address.  This is an SMN router that you're interacting with.

chipsec/modules/common/rom_armor.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def is_supported(self) -> bool:
+        return self.cs.is_amd()
+
+    def check_ROMAMOR(self) -> bool:
+        reg_value = self.cs.psp.smu_read32(SMU_PSP_SMN_BASE + SMU_PSP_MBOX_CMD_STATUS)
+        self.logger.log_information(f"PSP Mailbox Status 0x{reg_value:X}")
+        reg_value = self.cs.psp.smu_read32(SMU_PSP_SMN_BASE + 0x109fc)
+        self.logger.log_information(f"PSP Mailbox Features 0x{reg_value:X}")
+        hsti = self.cs.psp.query_HSTI()
```
superm1, line 58: A few FYIs.
1. There is no guarantee that the command exists on newer/all platforms.  The kernel only uses this command on platforms that don't report the security reporting capability bit.
2. There is a lot of other interesting information in this return.  The kernel code has offsets documented that you may want to look at.

chipsec/modules/common/smm_lock.py
```diff
@@ -0,0 +1,78 @@
...
+
+        return self.result.getReturnCode(res)
+
+    # --------------------------------------------------------------------------
+    # run( module_argv )
+    # Required function: run here all tests from this module
+    # --------------------------------------------------------------------------
+    def run(self, module_argv: List[str]) -> int:
+        self.logger.start_test("Compatible SMM Memory (SMRAM) Lock Protection")
+        return self.check_HWCR()
```
superm1, line 78: presumably you want a newline at the EOF.

chipsec/cfg/1022/strix.xml
npmitche: Currently we don't support adding cpuid to a sku tag, but I think this would be a great addition.

npmitche (commented):
Thank you for the updates!
If possible, please also clean up the unused imports to address to codeql items. 

If any of my comments are wrong or n/a, feel free to simply reply to them with why. =)

npmitche (approved):

Sae86 (approved):

chipsec/cfg/1022/strix.xml
```diff
@@ -0,0 +1,234 @@
+<?xml version="1.0"?>
+<configuration platform="Strix Point">
```
npmitche, line 2: The `<sku ... code="Strix" ...>` needs to match the `<configuration platform="Strix Point">`:
https://github.com/chipsec/chipsec/actions/runs/11599009964/job/32297207679?pr=2343

I'm contemplating removing the code attribute on the sku, because it's actually redundant and not needed, but per history, they currently need to match.

npmitche (changes requested):

npmitche (approved):

PR #2362, opened 2024-11-24
SMI enumeration using the standard deviation

cpey:
This introduces the standard deviation to identify long-running SMIs, requiring
less fine tunning when changing hardware. An outlier is detected when the new
measurement takes longer or shorter than the number of standard deviations
defined by the constant `OUTLIER_STD_DEV`, which is set to 2.

For this to work, a calibration is performed when a new SMI number is being
scanned.  The number of samples used in the calibration is defined by
`SCAN_CALIB_SAMPLES`, currently set to 50.

The other changes are related to how the average is calculated and the creation
of a new class, `smi_stats`, to handle the statistics of the current SMI and
the combined historical results.

### DIFF
--- a/chipsec/modules/tools/smm/smm_ptr.py
+++ b/chipsec/modules/tools/smm/smm_ptr.py
@@ -254,8 +254,6 @@ class scan_track:
         self.max = smi_info(0)
         self.min = smi_info(2**32 - 1)
         self.outlier = smi_info(0)
-        self.acc_smi_duration = 0
-        self.acc_smi_num = 0
         self.avg_smi_duration = 0
         self.avg_smi_num = 0
         self.outliers = 0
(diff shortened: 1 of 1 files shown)


npmitche (approved):
Does this add a noticeable overhead?

cpey:
> Does this add a noticeable overhead?

@npmitche, the impact varies depending on the scan configuration. Using the default values, with `FUZZ_SMI_FUNCTIONS_IN_ECX` set to True, `MAX_SMI_FUNCTIONS` set to 0x10, and `MAX_SMI_DATA` set to 0x100, a total of 4096 (0x10 * 0x100) SMIs would be executed per SMI code. Adding 50 SMIs for the initial calibration represents a 1.2% increase. The impact is more significant when `FUZZ_SMI_FUNCTIONS_IN_ECX` is set to False (which is not the default). However, in both cases, this impacts only the scan mode. Also, note that this impact is not noticeable unless you are scanning a large SMI address space.

dscott90 (approved):

PR #2405, opened 2025-02-04
Change module result to "information" if CPU is a sample part in sgx_check module

dscott90:


chipsec/modules/common/sgx_check.py
```diff
@@ -210,6 +210,7 @@ def check_sgx_config(self) -> int:
         debug_res, debug_result = self.check_debug()
         if self.cs.control.is_defined('SamplePart') and self.cs.control.get('SamplePart') == 1:
             self.logger.log_passed('CPU is a Sample Part. Test is N/A.')
```
Sae86, line 212: This line could be moved to run() in a new if statement that looks for ModuleResult.INFORMATION. The way it is now it will print a log_failed message even if the result is information.

Sae86 (changes requested):

npmitche (dismissed):

Sae86 (approved):

npmitche (approved):

PR #2413, opened 2025-02-14
fix for flake8 rule E231

WilliamLeara:
This is one in a series of PRs to get the source compliant with our .flake8 configuration.

This commit fixes all the E231 errors.  They were found by running "flake8 ." from the root directory.  E231 is an error defined by "pycodestyle", and has no functional impact to the source.  This is simply fixing whitespace as defined by CHIPSEC's .flake8 configuration.

background:
https://www.flake8rules.com/rules/E231.html

versions:  flake8 v7.1.1, python v3.12.6

### DIFF
--- a/chipsec/library/register.py
+++ b/chipsec/library/register.py
@@ -550,7 +550,7 @@ class Register:
         rtype = reg['type']
         reg_str = ''
         reg_width = reg["size"] * 2
-        reg_val_str = f'0x{reg_val:0{reg_width: d}X}'
+        reg_val_str = f'0x{reg_val:0{reg_width:d}X}'
         if RegisterType.PCICFG == rtype or RegisterType.MMCFG == rtype:
             if bus is not None:
                 b = bus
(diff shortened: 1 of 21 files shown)


npmitche (approved):

dscott90 (approved):

chipsec/library/register.py
```diff
@@ -550,7 +550,7 @@ def print(self, reg_name: str, reg_val: int, bus: Optional[int] = None, cpu_thre
         rtype = reg['type']
         reg_str = ''
         reg_width = reg["size"] * 2
-        reg_val_str = f'0x{reg_val:0{reg_width:d}X}'
+        reg_val_str = f'0x{reg_val:0{reg_width: d}X}'
```
npmitche, line 553: This change needs to be backed out. This one is in a format string and causes an error.

Everything else looks fine.

npmitche (changes requested):

chipsec/library/register.py
WilliamLeara: Profuse apologies for that, I've committed a fix, and GitHub has added it to the PR.  Thanks @npmitche.

npmitche (approved):

Issue #2253, opened 2024-08-06
Chipsec fails to detect VariableSmm EFI binaries within UEFI images


platomav:
Hello,

After the recent changes of #1790 and #2197 by @BrentHoltsclaw, I run a regression test on some UEFI image samples with known results based on earlier Chipsec versions.

In some cases/samples, I noticed that the latest Chipsec (e.g. [1.13.4](https://github.com/chipsec/chipsec/releases/tag/1.13.4)) misses some EFI binaries when scanned via "scan_image" module:

![Screenshot 2024-08-06 223348](https://github.com/user-attachments/assets/09ce0ef3-545a-4742-8b8f-aced6b6d779e)

**73d17a462c4999d1415259e15c2ded55457b06ac.BIN**

```json
[
    {
        "name": "VariableSmm",
        "guid": "23A089B3-EED5-4AC5-B2AB-43E3298C2343",
        "sha1": "064e939d8f30553bc71e151d5bee9fdcc199a04f",
        "type": "S_PE32"
    },
    {
        "name": "VariableSmm",
        "guid": "23A089B3-EED5-4AC5-B2AB-43E3298C2343",
        "sha1": "c61245ff69b4d1dea0accaf2bf54056bc95dc808",
        "type": "S_PE32"
    }
]
```

**78195fd9e70d6538b1873fa7b43d56c02036a66e.efi**

```json
[
    {
        "name": "VariableSmm",
        "guid": "23A089B3-EED5-4AC5-B2AB-43E3298C2343",
        "sha1": "b3c490762268d2d743cfa9ea0602826f9fe82d97",
        "type": "S_PE32"
    },
    {
        "name": "VariableSmm",
        "guid": "23A089B3-EED5-4AC5-B2AB-43E3298C2343",
        "sha1": "1240387225b71bbbda904ef341365414e7aab162",
        "type": "S_PE32"
    }
]
```

**BIOS_JHTC7_LN64_1.0.1.BIN**

```json
[
    {
        "name": "VariableSmm",
        "guid": "8307E5DF-33E5-4D68-96A3-ECA40C812D44",
        "sha1": "0d501c0b572bfe0ec73ed19cf13659978d628037",
        "type": "S_PE32"
    }
]
```

**R440-021400.cap**

```json
[
    {
        "name": "VariableSmm",
        "guid": "23A089B3-EED5-4AC5-B2AB-43E3298C2343",
[...]

Issue #2197, opened 2024-06-13
Chipsec fails to detect Computrace EFI binaries within UEFI image


platomav:
Hello,

As a continuation of https://github.com/chipsec/chipsec/issues/1790, I would like to focus on a specific sample which I know should contain at least 455 EFI binaries, while Chipsec is able to detect only 450:

`233baa0298ff7e9d46479868bbeb45948cbf649f_BIOS.cap`

Specifically, using the "scan_image" module, Chipsec detects 450 EFI binaries. There are at least 5 more which are not detected, related to "Computrace" and apparently stored within a **padding** section:

``` json
{
  "de4a46a4dbf04b40694687c5f3d826db265362973f6fc1740d54a46277d9d5f8": {
    "sha1": "59d7be2ead77825ad9f2a24ff8b93521c57fffc9",
    "guid": "8FEEECF1-BCFD-4A78-9231-4801566B3567",
    "name": "AbsoluteComputraceInstaller"
  },
  "399e5355ee4eed2331fefb3e49864da7d0d8c170ade413bea478a14469b540c2": {
    "sha1": "7e9d6cc0f5f28a020dbba47ecf07febbb0adb405",
    "guid": "4EFC51DA-23A6-4790-A292-4985C7F5CDEF",
    "name": "ComputraceEnablerDxe"
  },
  "e8ed47ff9d955bbb4fc349a7ecb5d5b8474cc20a3e32333f477c8bad433aa503": {
    "sha1": "cb06916d485b5dee6bcea0c70560539e8280edef",
    "guid": "4589CBF3-03F9-4998-9D6F-26343C69562A",
    "name": "ComputraceLoader"
  },
  "d77a47563aa18fcbdb92856cde52ea56531ee667b3dc91b09362ad67ca57e15c": {
    "sha1": "f190b200b1260cdbd3d9cbbdb4dcab9d929f4168",
    "guid": "18578E75-D073-4203-90D2-8788A871E44D",
    "name": "LenovoComputraceSmiServices"
  },
  "ef97bc7d42db1f91567f4dce4df19d783d7b7cb247c1c020067d428d50303f8e": {
    "sha1": "b3d4c94a06b1b5fefee629eaedcc94e4e9baa717",
    "guid": "AC60ED9F-523E-4F5B-94CA-3961346A00BA",
    "name": "LenovoVariableInitDxe"
  }
}
```

The area to look into should be:
- chipsec.chipsec.hal.spi_uefi
[...]

Issue #1790, opened 2023-05-10
Fallback parsing of detected UEFI volumes within padding/non-UEFI sections


platomav:
Hello,

There are various slightly odd UEFI capsules which do include valid UEFI volumes within areas which appear as Non-UEFI or padding:

![Screenshot 2023-05-10 202705](https://github.com/chipsec/chipsec/assets/11527726/17f11a36-50aa-4f40-8cee-12ec7cda482d)

Currently, Chipsec does not detect any EFI binaries within such (irregular ?) UEFI images when testing with "tools > uefi > scan_image" module, for example.

Would it be possible to add some form of "fallback" parsing in such cases to detect that valid UEFI volumes are included in non-expected sections and process them as normal, similar to what is shown at the picture being done by UEFITool?

Or maybe some older volume definition or similar is missing, perhaps?

![Screenshot 2023-05-10 204501](https://github.com/chipsec/chipsec/assets/11527726/72b368bb-ddff-4d14-9051-9377b2f47f9d)

I am attaching 14 sample capsule images, which should help to narrow this down.

[Non_UEFI.zip](https://github.com/chipsec/chipsec/files/11444699/Non_UEFI.zip)

Thanks!

BrentHoltsclaw:
@platomav can you help out with validation, I have a limited number of images to run against currently.

```[*] Found 165 EFI executables in UEFI firmware image '6d271ea0a7c4b0846404790fd9a7791bb3781be2_R101acer.Cap'```

platomav:
@BrentHoltsclaw Sure! Thanks for the work so far. I checked the PR against the provided images and saw the following:

*_oem --> Success (14) --> Appears to be correct, upon manual checking
*_R101acer.Cap --> Success (165) --> Appears to be correct, upon manual checking
*_SHW4.Cap --> Failure (0) --> Not sure exactly how many to expect, ~200 maybe (_FVH in both Non-UEFI and Padding)
*_FvNvStorage.fv --> Failure/Ignore (0) --> It appears to be NVRAM only, no EFI binaries (false positive)

Additionally, I found a much newer sample which works partially (450 instead of at least 455):

[233baa0298ff7e9d46479868bbeb45948cbf649f_BIOS.zip](https://github.com/chipsec/chipsec/files/11502298/233baa0298ff7e9d46479868bbeb45948cbf649f_BIOS.zip)

[...]

PR #2459, opened 2025-04-11
Added environment variable for brute force EFI module scan at HAL

platomav:
At HAL, when attempting to find EFI modules (VOLUME, FILE, SECTION), it is possible for some FILEs to be skipped because their contents happen to include valid-looking FVs inside. Since the "find_efi_modules" function was checking in succession for FV, FILE, SECTION but stopping at the first match (e.g. FV), Chipsec was effectively skipping actual FILEs that were mistaken for FVs instead.

sp101969

This is a known potential side effect of trying to detect EFI modules more broadly, especially when they are hidden/stored within Non-UEFI, Raw, Padding etc areas of an UEFI image. Meaning, instead of attempting to scan for FILEs only at a given blob, Chipsec tries to see if FV and/or SECTION can be found as well. The more broad EFI module scan approach is a good idea and has proven to help to detect weird EFI binaries within UEFI images:
- https://github.com/chipsec/chipsec/issues/1790
- https://github.com/chipsec/chipsec/issues/2197
- https://github.com/chipsec/chipsec/issues/2253

The solution to this is to have the option to attempt all of FV + FILE + SECTION without stopping at the first match. This has a couple side effects:
- Slightly higher processing time (usually within margin of error)
- Duplicate EFI module detections (efi trees at different depths)
  - https://github.com/chipsec/chipsec/issues/1790#issuecomment-1631297477
  - This can be handled with custom logic after the chipsec call, if one desires

To implement that, we could add an argument to the "find_efi_modules" function. However, because of the way that Chipsec HAL is laid out (bare functions), this is not really easy to do without a lot of variable transfers between functions. Ideally, in the future, these could be class methods and thus a class variable could easily control such behavior. For now, I decided to use an optional environment variable instead, which I named:
- CHIPSEC_HAL_FIND_EFI_MODULES_BRUTE_FORCE

Thus, if the environment variable "CHIPSEC_HAL_FIND_EFI_MODULES_BRUTE_FORCE" is present and non-empty, the HAL function "find_efi_modules" will not stop at the first successful call of FV or FILE or SECTION. Its output will return the EFI_MODULE list with all results from FV and/or FILE and/or SECTION detections.

In terms of samples, this FV within FILE (PE32 SECTION) behavior can be observed mostly at HP BIOS packages. For example (809 vs 811 EFI binaries detected, there are 2 being skipped):
- https://ftp.hp.com/pub/softpaq/sp101501-102000/sp101969.exe
- [S79_010101.zip](https://github.com/user-attachments/files/19708081/S79_010101.zip)

### DIFF
--- a/chipsec/hal/spi_uefi.py
+++ b/chipsec/hal/spi_uefi.py
@@ -340,21 +340,23 @@ def build_efi_tree(data: bytes, fwtype: Optional[str]) -> List['EFI_MODULE']:
 
 
 #
-# Attempts to find EFI modules by calling build_efi_tree (FV), build_efi_file_tree (FILE),
-# and build_efi_modules_tree (SECTION), in succession, stopping on first successful call.
+# Attempt to find EFI modules by calling build_efi_tree (FV), build_efi_file_tree (FILE),
+# and build_efi_modules_tree (SECTION), in succession. Stop on first successful call,
+# unless the environment variable CHIPSEC_HAL_FIND_EFI_MODULES_BRUTE_FORCE is defined.
 #
 def find_efi_modules(data: bytes, fwtype: Optional[str], polarity: bool, data_size: Optional[int] = None,
                      data_start: Optional[int] = None) -> List['EFI_MODULE']:
     data_len: int = len(data) if data_size is None else data_size
     data_off: int = 0 if data_start is None else data_start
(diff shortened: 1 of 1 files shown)


npmitche (approved):
I like this idea, though I would prefer to use the options.py functionality instead of an environment variable. Though we'll need to make some changes to add a `hal_options.ini` file I think, so we can do that after the fact.

dscott90 (approved):

PR #2487, opened 2025-05-06
dev/chipsec2 - cfg platform object structure usage update

npmitche:
Note: with how the registers are retrieved now, if the scope has eg `"BC": "8086.SPI.BC"` it will fail because it will be looking for `"8086.SPI.BC.BC"` which doesn't exist. So it will need to be `"BC": "8086.SPI"`

Also note, the scopes are now cleared after each module is executed.

### DIFF
--- a/chipsec/library/device.py
+++ b/chipsec/library/device.py
@@ -37,7 +37,7 @@ class Device:
         return None
         
     def get_list_by_name(self, device_name: str):
-        return self.cs.Cfg.platform.get_matches_from_scope(device_name)
+        return [ip.obj for ip in self.cs.Cfg.platform.get_matches_from_scope(device_name)]
 
     # def get_first_bus(self, device: dict) -> int:
     #     """Retrieves first value in bus list for PCI device"""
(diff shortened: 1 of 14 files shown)


chipsec/hal/common/mmio.py
```diff
@@ -191,6 +191,8 @@ def flush_bar_address_cache(self) -> None:
     # Get base address of MMIO range by MMIO BAR name
     #
     def get_MMIO_BAR_base_address(self, bar_name: str, instance: Optional['PCIObj'] = None) -> Tuple[int, int]:
+        # if bar_name == "8086.HOSTCTL.VTBAR":
```
BrentHoltsclaw, line 194: looks like debugging comments

chipsec/library/device.py
```diff
@@ -122,12 +123,12 @@ def get_bus(self, device_name: str) -> List[int]:
         #         logger().log_important(f"No bus value defined for device '{device_name}'")
         # return buses
 
-    def switch_def(self, target_device: str, source_device: str) -> None:
-        """Changes bus, device, and function values of PCI device"""
-        (b, d, f) = self.get_BDF(source_device)
-        self.cs.Cfg.CONFIG_PCI[target_device]['bus'] = b
-        self.cs.Cfg.CONFIG_PCI[target_device]['dev'] = d
-        self.cs.Cfg.CONFIG_PCI[target_device]['fun'] = f
+    # def switch_def(self, target_device: str, source_device: str) -> None:
```
BrentHoltsclaw, line 126: still needed?

chipsec/library/register.py
```diff
@@ -223,6 +238,7 @@ class BaseConfigRegisterHelper(BaseConfigHelper):
     def __init__(self, cfg_obj):
         super(BaseConfigRegisterHelper, self).__init__(cfg_obj)
         self.name = cfg_obj['name']
+        # self.full_name = cfg_obj['full_name']
```
BrentHoltsclaw, line 241: remove?

chipsec/modules/common/remap.py
```diff
@@ -103,7 +103,8 @@ def check_remap_config(self) -> int:
         # tmp = self.cs.Cfg.platform.get_matches_from_scope('8086.HOSTCTL.MCHBAR*')
         # tmp2 = ObjList(tmp[0].REMAPBASE + tmp[1].REMAPBASE)
         # tmp3 = self.cs.Cfg.platform.get_matches_from_scope('8086.HOSTCTL.MCHBAR*.REMAPBASE')
-        # tmp4 = self.cs.Cfg.platform.get_register_matches_from_scope('8086.HOSTCTL.MCHBAR*.REMAPBASE')
+        # tmp4 = self.cs.Cfg.get_objlist('8086.HOSTCTL.MCHBAR*.REMAPBASE')
```
BrentHoltsclaw, line 106: looks like the tmp vars can be cleaned up

BrentHoltsclaw (approved):

dscott90 (approved):

PR #2564, opened 2025-08-25
access_uefispec: accept both hex and base-10 formatted attr input

tklengyel:
Beside the functional improvement, converting the input to int is required depending on what Python version is being used to avoid the variable propagating down the call-chain as a str.

### DIFF
--- a/chipsec/modules/common/uefi/access_uefispec.py
+++ b/chipsec/modules/common/uefi/access_uefispec.py
@@ -257,7 +257,7 @@ class access_uefispec(BaseModule):
         attribute = None
 
         if len(module_argv) > 1:
-            attribute = module_argv[1]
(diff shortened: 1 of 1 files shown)


chipsec/modules/common/uefi/access_uefispec.py
```diff
@@ -257,7 +257,8 @@ def run(self, module_argv: List[str]) -> int:
         attribute = None
 
         if len(module_argv) > 1:
-            attribute = module_argv[1]
+            _attr = module_argv[1]
+            attribute = int(_attr, base=16) if _attr[:2] == "0x" else int(_attr)
```
npmitche, line 261: Please use `int(_attr, base=0)` As this will do the same thing. (Thanks for catching this one!)

npmitche (approved):

PR #2567, opened 2025-08-29
dev/chipsec2 fixes

BrentHoltsclaw:


chipsec/hal/intel/cpu.py
```diff
@@ -173,8 +173,8 @@ def get_TSEG(self) -> Tuple[int, int, int]:
         if self.cs.is_server():
             # tseg register has base and limit
             tseg_base = self.cs.register.get_list_by_name('8086.MEMMAP_VTD.TSEG').read_field('base', True)[0]
-            tseg_limit = self.cs.register.get_list_by_name('8086.MEMMAP_VTD.TSEG').read_field('limit', True)[0]
-            tseg_limit += 0xFFFFF
+            tseg_limit = self.cs.register.get_list_by_name('8086.MEMMAP_VTD.TSEG').read_field('limit', False)[0]
+            tseg_limit = (tseg_limit << 20) + 0xFFFFF
```
npmitche, line 177: I think if we left it the way it was, it keeps the bit shift in the xml, which I prefer. Can you try it and confirm?

tests/hal/test_cpu.py
```diff
@@ -238,7 +238,7 @@ def test_hal_cpu_get_TSEG_is_server_true_cmd(self):
         reg_call_1 = call('8086.MEMMAP_VTD.TSEG')
         reg_call_2 = call('8086.MEMMAP_VTD.TSEG')
         field_call_1 = call('base', True)
-        field_call_2 = call('limit', True)
```
npmitche, line 241: If you confirm that undoing the cpu hal change above, then this will need to be reverted too.

PR #2616, opened 2025-11-19
Implement SMBus block read/write

smx-smx:
Tested on Intel Z270

### DIFF
--- a/chipsec/hal/smbus.py
+++ b/chipsec/hal/smbus.py
@@ -45,6 +45,8 @@ SMBUS_POLL_COUNT = 1000
 SMBUS_COMMAND_WRITE = 0
 SMBUS_COMMAND_READ = 1
 
+SMBUS_BLOCK_DONE_STATUS = 0x80
+
 
 class SMBus(hal_base.HALBase):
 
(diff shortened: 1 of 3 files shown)


npmitche (changes requested):
None of `self.cs.*_register*()` function calls will work in the current version of chipsec (1.13.x) as they were refactored in this commit: https://github.com/chipsec/chipsec/commit/e569136c9b2fc8cbd560501caf62460cdc7969c4

Please be sure to update to and test it on the latest version of the `main` branch.

chipsec/hal/smbus.py
```diff
@@ -137,6 +149,7 @@ def _wait_for_cycle(self) -> bool:
             busy = self.cs.register.get_field(self.smb_reg_status, sts, 'BUSY')
             intr = self.cs.register.get_field(self.smb_reg_status, sts, 'INTR')
             failed = self.cs.register.get_field(self.smb_reg_status, sts, 'FAILED')
+            ds = self.cs.register.get_field(self.smb_reg_status, sts, 'DS')
```
npmitche: It looks like `ds` is not used in this function and does not need to be read.

npmitche:
Would also like to see the Unit tests passing. You can run them locally using `python3 -m pytest` from the root chipsec directory.

PR #2627, opened 2025-12-05
Update UEFIVar_Fuzz to send correct attribute type

dscott90:


chipsec/modules/tools/uefi/uefivar_fuzz.py
```diff
@@ -201,6 +201,11 @@ def run(self, module_argv):
                     else:
                         attrib_prim = prim.dword(value=_ATTRIB)
                         _ATTRIB = attrib_prim.render()
```
npmitche: Instead of doing this calculation, you can just set _ATTRIB to `attrib_prim.value` (instead of calling `render()`) here and on line 200.

PR #2678, opened 2026-02-11
Update configs and files for some external tests

npmitche:


chipsec-bbci (dismissed):

chipsec/cfg/parsers/ip/platform.py
```diff
@@ -629,24 +629,24 @@ def __init__(self, name: str, barobj: Any) -> None:
         self.name = name
         self.obj = barobj
 
-    def _get_next_level_list(self) -> List[str]:
-        """Get list of register names."""
-        return list(self.register_list.keys())
+    # def _get_next_level_list(self) -> List[str]:
```
BrentHoltsclaw: If the code is commented no longer needed, why not delete?

chipsec/library/register.py
```diff
@@ -337,12 +278,10 @@ def has_field(self, reg_name: str, field_name: str) -> bool:
             reg_defs = self.cs.Cfg.get_reglist(reg_name)
         except RegisterNotFoundError:
             return False
-        for reg_def in reg_defs:
-            try:
-                return field_name in reg_def.fields
-            except KeyError:
-                return False
-        return False
+        try:
```
BrentHoltsclaw, line 281: Is a behavioral change ok and not going to break anything?

npmitche: I hit an instance where a register pulled two different register definitions and one had the field and the other didn't. So, the "has_field" check passed, but when it went to read the field, it threw an error because the second one didn't have the field. 
I updated the test to run "filter_with_fields" before just blindly reading, but figured this change would also be good. 

I don't think it will break anything. If you see something, please let me know.

BrentHoltsclaw (approved):

PR #2687, opened 2026-02-23
Class cleanup

BrentHoltsclaw:
Changes to help with mmio memory ranges

### DIFF
--- a/chipsec/cfg/parsers/ip/platform.py
+++ b/chipsec/cfg/parsers/ip/platform.py
@@ -233,6 +233,21 @@ class Platform(Recursable):
         self.vendor_list.remove(vendor.name)
         self.__delattr__(f'_{vendor.name}')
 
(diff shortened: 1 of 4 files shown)


chipsec/cfg/parsers/ip/platform.py
```diff
@@ -608,7 +641,7 @@ def _get_next_level(self, bar_id: str) -> 'Bar':
             raise PlatformConfigError(f'Next Level: {bar_id} not found in IP: {self.name}')
 
 
-class Bar(RegisterList):
+class Bar(Recursable, RegisterList):
```
npmitche: What is the reason for this being added?

npmitche (approved):
@BrentHoltsclaw  can you test and make sure this satisfies what you need it to do?

dscott90 (approved):

PR #2692, opened 2026-02-26
UEFI decode: fix parsing bugs and add section decoders

BrentHoltsclaw:
Fix multiple bugs in the UEFI firmware image decode/modify pipeline and add
missing section-type decoders per the PI and UEFI specifications.

Rebased onto current `chipsec2`.

- Correct the DEPEX opcode table to PI 1.9 (Vol II §10.7 DXE, Vol I §5.7.1.1 PEI).
  The previous table matched no PI encoding, so every dependency expression
  decoded to the wrong opcodes.
- Correct `EFI_SECTION_DISPOSABLE` from `0x1A` to `0x03`. `0x1A` is not an
  assigned section type. Since `0x03` is now recognised it no longer reaches the
  unknown-section rescan path, so it gets an explicit handler that recurses into
  it as an encapsulation section.
- Add a length guard before unpacking `EFI_COMPRESSION_SECTION`; a truncated
  compression section previously raised `struct.error` out of the pipeline.
- Restore the `decode_uefi_region()` return-value check so `uefi decode` sets
  `ExitCode.ERROR` on failure.
- Restore the `parse_uefi_region_from_file()` signature: returns the parsed tree,
  and `filetype` no longer uses a mutable default.

## Additional fixes found while addressing the review

- FFS large-file header checksum covered a hardcoded zero instead of the real
  `ExtendedSize`, and the file checksum started after the 24-byte header rather
  than the full 32-byte one. Behaviour for normal 24-byte headers is unchanged.
- The inter-file gap scan sliced `[padding:offset - 1]`, dropping the gap's final
  byte. A firmware volume occupying a gap exactly then failed its length check and
  was misparsed into unrelated fragments instead of being reported as one volume.
  The padding-file case also built its section from the region *after* the pad
  while deriving children from the pad itself.
- `modify_uefi_region()` no longer rewrites `FvLength`. The size delta is absorbed
  by the free space at the end of the FV, so the volume keeps its original length;
  rewriting it left the header inconsistent with the image and made the output
  unparseable by `NextFwVolume()`.
- TE image header parsing used a redundant unpack and a comment describing a
  layout that does not exist (real layout: PI Vol I §15).

## Decoding stays deliberately permissive

Firmware in the wild is not always conformant, and this pipeline exists partly to
find content that vendors hide or misplace. Opcodes PEI does not define are decoded
and tagged `!non-PEI` rather than dropped, an unknown opcode reports how many bytes
were left undecoded, and non-blank data after `END` is reported — so payload hidden
in a dependency section stays visible.

## Bug fixes

- Fix broken imports in `spi.py` after the package restructure
- Fix `variables={}` reset inside loop in `varstore.py` (lost all but last store)
- Fix hardcoded `'nvar'` type in `hal/common/uefi.py` (should use `self._FWType`)
- Fix large-file `cur_offset` calculation in `NextFwFile`
- Fix GUID type mismatch in modify commands (string vs UUID)
- Fix wrong LZMA compression constant in assemble
- Fix erroneous `break` in `uefi_cmd.py` filetype processing loop
- Fix `self.rom_file` AttributeError in modify commands

## New section decoders

- DEPEX (PEI/DXE/MM dependency expression opcodes)
- VERSION (BuildNumber + version string)
- FREEFORM_SUBTYPE_GUID (SubTypeGuid extraction)
- `EFI_SECTION_COMPRESSION` CompressionType field per PI Vol III

## New capabilities

- Strip UEFI capsule headers (.cap files) before FV parsing
- Detect Intel CPU microcode blobs in RAW sections
- Label ACPI tables in RAW sections using hal/common/acpi signatures
- Parse PE32/PE32+/TE executable metadata (machine, subsystem, entry point)
- Parse FV Extended Header with FvNameGuid
- Handle vendor GUID-defined sections (AMI signed, Phoenix) with decompress fallback
- Attempt FFS parse on FVs with unrecognized GUIDs
- Decode PEI/DXE Apriori file dispatch order
- FTW working/spare block detection and labeling

## Files changed

- `chipsec/library/uefi/fv.py` — FV/FFS/section parsing and assembly
- `chipsec/library/uefi/spi.py` — decode pipeline orchestration
- `chipsec/library/uefi/common.py` — EFI_GUID_FMT helper
- `chipsec/library/uefi/__init__.py` — license header
- `chipsec/hal/common/uefi.py` — FWType and NVRAM fixes
- `chipsec/utilcmd/uefi_cmd.py` — CLI command fixes

Tested against multiple real SPI flash images.


dscott90 (approved):

dscott90 (dismissed):

npmitche:
@platomav Can you take a look at this, #2693 and #2694?

chipsec/library/uefi/compression.py
```diff
@@ -86,9 +89,11 @@
     COMPRESSION_TYPE_TIANO,
     COMPRESSION_TYPE_UEFI,
     COMPRESSION_TYPE_LZMA,
+    COMPRESSION_TYPE_LZMAF86,
```
platomav: There is no way to tell LZMA from LZMAF86 apart. They are both LZMA compressed. The latter simply requires an extra step after decompression (or similarly an extra step before compression). So unless you have the specific GUID of LZMAF86, you cannot know if you need the x86 transformation/filter or not. I assume this list includes compression schemes to blindly test, so no GUID-defined hint, thus there is no reason to include LZMA and LZMAF86, you might as well leave LZMA only and hope it's not LZMAF86. If it's LZMAF86, you'll just get corrupted data.

chipsec/library/uefi/compression.py
```diff
@@ -86,9 +89,11 @@
     COMPRESSION_TYPE_TIANO,
     COMPRESSION_TYPE_UEFI,
     COMPRESSION_TYPE_LZMA,
+    COMPRESSION_TYPE_LZMAF86,
     COMPRESSION_TYPE_ZLIB_AMD,
     COMPRESSION_TYPE_GZIP,
-    COMPRESSION_TYPE_BROTLI
+    COMPRESSION_TYPE_BROTLI,
+    COMPRESSION_TYPE_ZSTD
```
platomav: Where is ZSTD mentioned to be a supported algorithm? Maybe only a specific vendor is using it on the side?

chipsec/library/uefi/compression.py
```diff
@@ -178,6 +183,12 @@ def decompress_efi_binary(self, compressed_data: bytes, compression_type: int) -
                     raise ValueError('Could not validate decompressed BROTLI data length')
             except Exception as error:
                 logger().log_hal(f'Cannot decompress BROTLI data: {error}')
+        elif compression_type == COMPRESSION_TYPE_ZSTD and modules.get('zstandard', False):
```
platomav: This appears to require an external dependency named "zstandard" which is not referenced at the project requirements. Also, fyi, Python 3.14 now has zstd support as part of the standard library and, according to the [PEP](https://peps.python.org/pep-0784/), it was based on https://github.com/Rogdham/pyzstd project so maybe that could be used instead to be forward compatible.

chipsec/library/uefi/fv.py
```diff
@@ -168,11 +204,95 @@
 EFI_GUIDED_SECTION_TIANO = UUID('A31280AD-481E-41B6-95E8-127F4C984779')
 EFI_GUIDED_SECTION_ZLIB_AMD1 = UUID('CE3233F5-2CD6-4D87-9152-4A238BB6D1C4')
 EFI_GUIDED_SECTION_ZLIB_AMD2 = UUID('991EFAC0-E260-416B-A4B8-3B153072B804')
+EFI_GUIDED_SECTION_ZSTD = UUID('5D4A891A-7EB4-4343-A93E-8A06C3A0F5C3')
```
platomav: Cannot find any reference to this GUID-defined section. Which vendor uses this? Can you provide a sample perhaps?

platomav (commented):
This MR is rather large so it's difficult to review, and I don't have a testing flow for Chipsec 2, so I just took a look. I left some comments regarding the compression module, since I had worked on it a bit in the past.

chipsec/library/uefi/compression.py
BrentHoltsclaw: Good point I missed that detail when porting.  Thanks

chipsec/library/uefi/compression.py
BrentHoltsclaw: Apologies, AI hallucination, removing all instances.

platomav:

# PR #2692 Review Comment

Please address the following before merge. These are not just style issues; several of them are spec mismatches or regressions that can produce incorrect decode output or corrupt modified firmware images.

## Requested fixes

1. Correct the DEPEX opcode table in `chipsec/library/uefi/fv.py`.

   The new `decode_depex()` table does not match the PI specification. The current code maps `PUSH=0x00`, `TRUE=0x04`, `END=0x06`, `BEFORE=0x08`, and `AFTER=0x09`. For DXE DEPEX, PI defines:

   - `0x00 BEFORE`
   - `0x01 AFTER`
   - `0x02 PUSH`
   - `0x03 AND`
   - `0x04 OR`
   - `0x05 NOT`
   - `0x06 TRUE`
   - `0x07 FALSE`
   - `0x08 END`
   - `0x09 SOR`

   PEI DEPEX also starts `PUSH` at `0x02` and uses the same boolean opcode values. Please update the constants and decoder accordingly. If PEI/DXE/MM need different opcode handling, the decoder should be section-type aware instead of using one incorrect universal table.

   Ground truth references:

   - PI 1.9, Volume II, DXE Dispatcher, "Dependency Expression Opcode Summary": https://uefi.org/specs/PI/1.9/V2_DXE_Dispatcher.html
   - PI 1.9, Volume I, PEI Foundation, "Dependency Expression Opcode Summary": https://uefi.org/specs/PI/1.9/V1_PEI_Foundation.html

   Quick repro against this PR stack:

   ```python
   from uuid import UUID
   from chipsec.library.uefi.fv import decode_depex

   print(decode_depex(bytes([0x06, 0x08])))
   # Expected: TRUE END
   # Current:  END

   g = UUID("cccccccc-cccc-cccc-cccc-cccccccccccc")
   print(decode_depex(bytes([0x00]) + g.bytes_le))
   # Expected for DXE: BEFORE {cccccccc-cccc-cccc-cccc-cccccccccccc}
   # Current:          PUSH {cccccccc-cccc-cccc-cccc-cccccccccccc}
   ```

2. Correct `EFI_SECTION_DISPOSABLE`.

   `EFI_SECTION_DISPOSABLE` is currently added as `0x1A`. PI code definitions define `EFI_SECTION_DISPOSABLE` as `0x03`; `0x1A` is not the disposable section type. Please update both the constant and `SECTION_NAMES`, and leave `0x1A` as unknown/reserved unless there is a vendor-specific reason to label it.

   Ground truth reference:

   - PI 1.9, Volume III, Code Definitions: https://uefi.org/specs/PI/1.9/V3_Code_Definitions.html

   Quick repro against this PR stack:

   ```python
   from chipsec.library.uefi.fv import NextFwFileSection

   sec = (8).to_bytes(3, "little") + b"\x03" + b"abcd"
   print(NextFwFileSection(sec, len(sec), 0, False).Name)
   # Expected: S_DISPOSABLE
   # Current:  S_UNKNOWN_03

   sec = (8).to_bytes(3, "little") + b"\x1a" + b"abcd"
   print(NextFwFileSection(sec, len(sec), 0, False).Name)
   # Expected: unknown/reserved
   # Current:  S_DISPOSABLE
   ```

3. Add a length guard for compression section header parsing.

   The new PI-directed compression parsing unconditionally unpacks `EFI_COMPRESSION_SECTION` from `sec.Image[sec.HeaderSize:sec.HeaderSize + EFI_COMPRESSION_SECTION_size]`. A malformed/truncated compression section can now raise `struct.error` instead of being handled as malformed input.

   Please check that `len(sec.Image) >= sec.HeaderSize + EFI_COMPRESSION_SECTION_size` before unpacking, log/comment the malformed section, and continue safely. Please add a regression test for this case.

   Quick repro:

   ```python
   from chipsec.library.uefi.spi import build_efi_modules_tree
   from chipsec.library.uefi.fv import EFI_SECTION_COMPRESSION

   malformed = (8).to_bytes(3, "little") + bytes([EFI_SECTION_COMPRESSION]) + b"\x00" * 4
   build_efi_modules_tree(None, malformed, len(malformed), 0, False)
   # Current: struct.error: unpack requires a buffer of 5 bytes
   ```

4. Fix FV modification header recalculation.

   `modify_uefi_region()` now rewrites `FvLength` as `OldFvLength + FvLengthChange`, but the surrounding code still keeps the output image/FV padded back to the original size. That can make the FV header inconsistent with the actual image and block map.

   Please either:

   - keep `FvLength` unchanged when insertion/removal is absorbed by padding, or
   - update the FV image size and block map consistently if the FV really changes size.

   Please add tests for remove, insert-before, insert-after, and replace with different-sized files, and validate that `NextFwVolume()` can parse the modified output.

   Quick repro:

   ```text
   Inserting into an 0x8000 FV produced a modified image still 0x8000 bytes long,
   but rewrote FvLength to 0x8030. NextFwVolume() then failed to parse it.
   ```

5. Restore `uefi decode` failure propagation.

   This PR removes the `if not decode_uefi_region(...): self.ExitCode = ExitCode.ERROR` check from `chipsec/utilcmd/uefi_cmd.py`. That is part of the reason the final stack reports success for non-firmware input. Please keep the decode return-value check and set `ExitCode.ERROR` on failure.

   Existing test that catches this:

   ```bash
   ```

6. Clean up the stale Zstd claim in the PR description/commit text.

   The current code no longer contains `zstd`/`zstandard` implementation, tests, or dependency changes. If the earlier Zstd implementation was intentionally removed, please remove the remaining description claim so future reviewers do not look for nonexistent functionality.

npmitche (approved):

PR #2693, opened 2026-02-26
UEFI decode: output improvements, crash fixes, and code quality

BrentHoltsclaw:
Depends on #2692  (UEFI decode: fix parsing bugs and add section decoders).

Improve decode output handling, fix crash bugs, and clean up code quality
across the UEFI decode pipeline.

## Output improvements
- Replace `set_log_file()` redirect pattern with direct file I/O for `.lst`
  artifacts — the logger stays on its normal handlers throughout decode,
  eliminating log file conflicts where NVRAM output clobbered the `.UEFI.lst`
  handler mid-flow
- Unify `save_efi_tree` / `save_efi_tree_filetype` into a single function
  with an optional `filetype` parameter (removes ~60 lines of duplicate code)
- Add offset and GUID context to parse warning messages so users can trace
  which module triggered each warning

## Bug fixes
- `identify_EFI_NVRAM`: return `None` instead of `''` on failure so callers'
  `if fwtype is None:` checks work correctly (was falling through to KeyError)
- `EFI_STATUS_DICT`: use `.get()` with fallback in var_write/var_delete to
  avoid KeyError on unexpected firmware status codes

## Code quality
- `NextFwVolume`: search bytes directly with `buffer.find(b'_FVH')` instead
  of converting the entire firmware image to string on every iteration
- `FvSum8`/`FvSum16`: iterate bytes directly instead of `ord()` conversion
- Bare `except` → `except Exception as e` with `repr(e)` logging
- Deduplicate `print_efi_variable` / `print_sorted_EFI_variables` formatting
- Convert `.format()` strings to f-strings in `uefi_cmd.py`

## Cleanup
- Remove unused `DecodeSection` stub and `section_dir_path` parameter
- Remove commented-out Python 2 `getEFIvariables_UEFI_Ex`
- Make `WRITE_ALL_HASHES` configurable via `CHIPSEC_WRITE_ALL_HASHES` env var
- Add docstrings to core public functions in fv.py, spi.py, varstore.py

## Files changed
- `chipsec/library/uefi/spi.py` — lst output, save_efi_tree unification
- `chipsec/library/uefi/varstore.py` — NVRAM return type, signature parsers, lst_lines
- `chipsec/library/uefi/fv.py` — dead code removal, checksum iteration
- `chipsec/utilcmd/uefi_cmd.py` — f-strings, set_log_file removal, crash fix

- **Restore the decode return contract.** `parse_uefi_region_from_file()` built the
  module tree but never returned it, and `decode_uefi_region()` fell out of the
  NVRAM path without returning at all. `uefi decode` therefore reported failure for
  every image, including valid firmware. The existing tests only assert the failure
  cases, so they passed while this was broken; verified separately that a
  well-formed FV now decodes with a `True` result and non-firmware still returns
  `False`.
- `uefi decode` sets `ExitCode.ERROR` again for both a missing input file and a
  decode failure.
- **Correct `EFI_CERT_X509_SHA256/384/512` parsing.** These carry the SHA hash of
  the TBS certificate followed by an `EFI_TIME TimeOfRevocation`, not a DER
  certificate, so the trailing bytes are no longer passed to the X.509 parser —
  which made valid dbx revocation entries log X.509 parse errors and discarded the
  timestamp. The fixed `SignatureSize` values in the signature type table
  (`0x40`/`0x50`/`0x60`) confirm the layout: 16-byte owner GUID + hash + 16-byte
  `EFI_TIME`. An all-zero timestamp is reported as unspecified, and trailing bytes
  beyond `EFI_TIME` are reported rather than dropped so non-conforming entries stay
  visible.
- Drop the unused `print_buffer_bytes` and `rsa_mod` imports flagged by CodeQL.


dscott90 (dismissed):

platomav:

# PR #2693 Review Comment

Please address the following before merge. This PR introduces the remaining decode return-value regression and adds Secure Boot signature parsing that does not match the UEFI-defined structure for X509 hash revocation entries.

## Requested fixes

1. Restore `ExitCode.ERROR` for missing input files.

   `chipsec/utilcmd/uefi_cmd.py` now logs a missing decode input file and returns without setting `self.ExitCode = ExitCode.ERROR`. Existing util tests expect this to be an error, and the current final stack returns `0` for a missing file.

   Requested fix:

   ```python
   if not os.path.exists(self.filename):
       self.logger.log_error(f"Could not find file '{self.filename}'")
       self.ExitCode = ExitCode.ERROR
       return
   ```

2. Restore the `parse_uefi_region_from_file()` return contract and `decode_uefi_region()` boolean result.

   `parse_uefi_region_from_file()` now returns `None` even after it successfully builds `tree`. `decode_uefi_region()` then does `tree = parse_uefi_region_from_file(...)`, so `bool(tree)` is always false for the FV decode path. At the end of the NVRAM path, `decode_uefi_region()` also falls through with no explicit return.

   Requested fix:

   - `parse_uefi_region_from_file()` should return the parsed tree.
   - `decode_uefi_region()` should return a boolean in all paths.
   - The `uefi decode` command should check that boolean and set `ExitCode.ERROR` on failure.
   - Avoid mutable defaults such as `filetype: List[int] = []`; use `Optional[List[int]] = None`.

   Existing failing test:

   ```bash
   ```

   Current failures:

   ```text
   test_decode_missing_file_returns_error: 0 != 16
   test_decode_non_firmware_file_returns_error: 0 != 16
   ```

3. Correct `EFI_CERT_X509_SHA256/384/512` parsing.

   The new `parse_x509_sha256()`, `parse_x509_sha384()`, and `parse_x509_sha512()` functions treat bytes after the hash as a DER X.509 certificate and call `parse_x509()`. UEFI defines these Secure Boot signature data formats as:

   - TBS hash
   - `EFI_TIME TimeOfRevocation`

   The data after the hash is not a DER certificate. Valid dbx revocation entries can therefore be logged as X.509 parse errors, and the revocation timestamp is ignored.

   Requested fix:

   - Parse the trailing bytes as `EFI_TIME TimeOfRevocation`.
   - Do not call `parse_x509()` for `EFI_CERT_X509_SHA256`, `EFI_CERT_X509_SHA384`, or `EFI_CERT_X509_SHA512`.
   - Add tests with owner GUID + hash + `EFI_TIME` entries for all three formats.

   Ground truth reference:

   - UEFI 2.11, Secure Boot and Driver Signing: https://uefi.org/specs/UEFI/2.11/32_Secure_Boot_and_Driver_Signing.html

4. Coordinate this PR with PR #2692's decode fix.

   PR #2692 already removed the `decode_uefi_region()` failure check in `uefi_cmd.py`. This PR then removes the missing-file error path and breaks the lower-level return contract. The final stacked behavior should be fixed in one place, but both PRs currently contribute to the regression.

npmitche (dismissed):

npmitche (approved):

PR #2694, opened 2026-02-26
UEFI decode: add unit tests for fv.py and spi.py

BrentHoltsclaw:
Depends on #2693 (UEFI decode: output improvements, crash fixes, and code quality).

Add comprehensive unit tests for the UEFI firmware volume parsing and decode
pipeline. 702 total tests all passing.

### tests/library/test_fv.py (89 tests)
- **Checksums**: FvChecksum8/FvChecksum16 correctness, overflow, zero, odd-length
- **NextFwVolume**: find single/multiple FVs, offset handling, extended headers,
  erase polarity, checksum validation, truncated/empty buffers, skip invalid candidates
- **NextFwFile**: find files, body content, FFS checksum, empty FV, corrupt header
  break behavior, blank gap handling
- **NextFwFileSection**: single/multiple sections, body content, UI strings,
  unknown types, empty data, invalid section skip, zero-size skip
- **GetFvHeader**: basic header, empty/truncated buffers
- **ValidateFwVolumeHeader**: valid, bad FvLength, bad HeaderLength, nonzero reserved
- **Assembly functions**: raw/file/section round-trips, alignment, compression,
  get_guid_bin
- **decode_depex**: all opcodes (PUSH, AND, OR, NOT, TRUE, FALSE, SOR, BEFORE,
  AFTER), truncated input, unknown opcodes
- **EFI module classes**: EFI_FV/EFI_FILE/EFI_SECTION str output, calc_hashes,
  FILE_TYPE_NAMES/SECTION_NAMES coverage, OEM type range
- **Regression tests**: corrupt header breaks scanning (not continues), blank gaps
  stop iteration, invalid sections skipped, invalid FV candidates skipped

### tests/library/test_spi.py (48 tests)
- **build_efi_tree**: single/multiple FVs, NVRAM detection, FTW blocks,
  unknown GUID fallback, empty data
- **build_efi_file_tree**: single file, RAW files, NVAR detection, padding,
  non-UEFI data tail
- **build_efi_modules_tree**: PE32/TE/RAW/UI/compression/FV image sections,
  unknown section types
- **update_efi_tree**: GUID propagation, UI string propagation
- **save_efi_tree**: JSON output structure, lst_lines accumulation, NVRAM extraction
- **modify_uefi_region**: remove/insert/replace operations
- **strip_capsule_header**: valid/invalid/nested capsules, non-capsule passthrough
- **Search helpers**: EFIModuleType, search_efi_tree, FILENAME generation

### tests/library/test_varstore.py (fixes)
- Fix test_run_placeholder_functions to use realistic byte data matching
  the updated parse_* function signatures

### tests/library/__init__.py
- Added to enable unittest discovery of library test modules

The dependency expression tests encoded the same incorrect opcode table as the
decoder, so they locked in the bug instead of catching it. They now assert the PI
values and check the constants directly. Also added:

- `EFI_SECTION_DISPOSABLE` is `0x03` and `0x1A` is not an assigned type, plus
  coverage that content nested in a disposable section is still decoded
- malformed compression section regression test (previously `struct.error`)
- FV modification tests for remove, insert-before, insert-after and replace with
  both larger and smaller files, asserting the image size and `FvLength` are
  unchanged and that `NextFwVolume()` can still parse the result. The previous
  tests only checked length and that something parsed, which is why they missed
  the `FvLength` corruption.
- `EFI_CERT_X509_SHA256/384/512` hash + `EFI_TIME TimeOfRevocation` coverage
- `EFI_FFS_FILE_HEADER2` checksum coverage
- padding-scan coverage: a gap section must describe the gap it reports, and a
  volume occupying a gap exactly must parse as a single volume

**Verified against the unfixed code:** these tests produce 18 failures and 4 errors
there, and pass only with the corresponding fixes applied.

### Note on `tests/library/__init__.py`

`tests/library` had no `__init__.py`, so `unittest discover` skipped the entire
directory. Adding it takes discovery from **418 to 702 tests**, bringing the
existing library tests into CI alongside the new ones.


dscott90 (approved):

dscott90 (dismissed):

platomav:

# PR #2694 Review Comment

Please update the tests before merge. This PR adds useful coverage, but some of the new tests currently lock in incorrect behavior from the earlier PRs instead of validating the UEFI/PI ground truth.

## Requested fixes

1. Update `decode_depex` tests to use the PI-defined opcode table.

   The tests in `tests/library/test_fv.py` currently encode the same wrong DEPEX opcode table added in PR #2692. For example, `test_true_end()` uses `bytes([0x04, 0x06])` as `TRUE END`, but official PI uses `0x06 TRUE` and `0x08 END`.

   Please update the tests so they fail against the current implementation and pass only after PR #2692 fixes the decoder.

   Suggested test cases:

   ```python
   def test_true_end(self):
       self.assertEqual(decode_depex(bytes([0x06, 0x08])), "TRUE END")

   def test_false_end(self):
       self.assertEqual(decode_depex(bytes([0x07, 0x08])), "FALSE END")

   def test_push_and_end(self):
       guid = UUID("AABBCCDD-1122-3344-5566-778899001122")
       data = bytes([0x02]) + guid.bytes_le + bytes([0x08])
       result = decode_depex(data)
       self.assertIn("PUSH", result)
       self.assertIn(str(guid), result.lower())
       self.assertIn("END", result)

   def test_dxe_before(self):
       guid = UUID("CCCCCCCC-CCCC-CCCC-CCCC-CCCCCCCCCCCC")
       data = bytes([0x00]) + guid.bytes_le + bytes([0x08])
       result = decode_depex(data)
       self.assertIn("BEFORE", result)
       self.assertIn(str(guid), result.lower())
       self.assertIn("END", result)
   ```

   Ground truth references:

   - PI 1.9, Volume II, DXE Dispatcher, "Dependency Expression Opcode Summary": https://uefi.org/specs/PI/1.9/V2_DXE_Dispatcher.html
   - PI 1.9, Volume I, PEI Foundation, "Dependency Expression Opcode Summary": https://uefi.org/specs/PI/1.9/V1_PEI_Foundation.html

2. Add a test for the `EFI_SECTION_DISPOSABLE` value.

   PR #2692 currently maps `EFI_SECTION_DISPOSABLE` to `0x1A`, but PI defines it as `0x03`. Please add a unit test that asserts a section type of `0x03` is named disposable and `0x1A` is not.

   Suggested assertion:

   ```python
   sec = _make_section(0x03, b"abcd")
   parsed = NextFwFileSection(sec, len(sec), 0, False)
   self.assertEqual(parsed.Name, "S_DISPOSABLE")
   ```

3. Add malformed compression-section coverage.

   PR #2692 can currently raise `struct.error` when a compression section has a valid common header but fewer than 5 bytes of compression payload header. Please add a regression test that ensures this does not throw and that the parser returns a safe malformed/empty result.

   Suggested fixture:

   ```python
   malformed = (8).to_bytes(3, "little") + bytes([EFI_SECTION_COMPRESSION]) + b"\x00" * 4
   build_efi_modules_tree(None, malformed, len(malformed), 0, False)
   ```

4. Add FV modification parseability tests for size-changing operations.

   The current `modify_uefi_region()` tests cover remove and same-size replace, but they do not catch the `FvLength` corruption from insert or different-size replace. Please add tests that:

   - insert a file before an existing file
   - insert a file after an existing file
   - replace with a different-sized file
   - remove a file
   - parse the result with `NextFwVolume()` or `build_efi_tree()`
   - assert the FV header length remains consistent with the output image and block map

5. Add Secure Boot X509_SHA tests once PR #2693 is fixed.

   PR #2693 should parse `EFI_CERT_X509_SHA256/384/512` as hash + `EFI_TIME TimeOfRevocation`, not hash + DER certificate. Please add tests with realistic signature data for all three formats so the parser does not regress back to treating the timestamp as DER.

6. Add or keep util command tests that catch decode failure returns.

   Existing `tests.utilcmd.uefi_cmd.test_uefi_cmd` fails against the current stack because `uefi decode` returns success for missing and non-firmware inputs. Please make sure the final stacked PR set includes these tests and that they pass:

   ```bash
   ```

npmitche (approved):

PR #2699, opened 2026-03-05
Resolve import errors for lib/uefi/spi.py

npmitche:


npmitche:
Already fixed in #2684

PR #2792, opened 2026-07-08
Update common.smrr module

dscott90:


tests/modules/test_smrr.py
```diff
@@ -0,0 +1,196 @@
...
+        module_self.cs.hals.cpu.check_SMRR_supported.return_value = False

+
+        result = smrr.check_SMRR(module_self, False)

+

+        module_self.result.setStatusBit.assert_called_once_with(module_self.result.status.NOT_APPLICABLE)

+        self.assertIn(result, (ModuleResult.PASSED, ModuleResult.NOTAPPLICABLE))
```
npmitche: Why are there two options for the result in this test?

npmitche (approved):

PR #2793, opened 2026-07-09
Add bios_version module

dscott90:


dscott90:
Redundant because firmware_info module already exists

PR #2844, opened 2026-08-11
Fix logger color arg causing TypeError in external Python logging han…

staslyakhov:
Cherry-pick logging fix to chipsec1. 

This bug is causing some issues on the nixpkgs build: https://github.com/NixOS/nixpkgs/issues/550494.

If we don't want to merge anything to chipsec1 anymore, we can also drop the chipsec1 package on nixpkgs entirely.


dscott90 (approved):

dscott90:
No problem merging this change into old chipsec. There will not be a formal release on 1.x.x with this fix.

